PR #1, opened 2018-05-03
Cans 155 init react app

jonbwalker:
Initializes the cans react app
run with yarn start
http://localhost:3000/


.gitignore
```diff
@@ -30,3 +30,9 @@
 /config/master.key
 
 .env
+
+# IDE
+*.idea
+
+# OS
+.DS_Store
```
edgibbs: Need to clean up a few newlines, guessing need to update an editor setting.

jonbwalker: :heavy_check_mark: done! 656fb93

PR #9, opened 2018-05-11
add cancel, save and submit buttons to form

jonbwalker:
these buttons have no functionality at this stage of the project


client/src/style.css
```diff
@@ -11,3 +11,12 @@
   color: #086D87;
   font-weight: 600;
 }
+
+/**********************
+  Domains
+***********************/
+.form-footer {
```
edgibbs: Just wondering if we should move this to SASS.

jonbwalker: Yes, let's bring in SASS

PR #21, opened 2018-05-18
Startup scripts and doco

jonbwalker:
* added commands to run build and bundle for the react app 
* added prettierignore
* removed port proxy - not needed
* updated readme


client/.prettierrc.json
```diff
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "es5"
-}
+  "trailingComma": "es5",
+  "printWidth": 120
+}
```
denys-davydov: The unknown symbol at the line end, possibly should be removed

README.md
```diff
@@ -22,3 +22,33 @@ Things you may want to cover:
...
+
+If this is your first time running the app:
+
+```yarn```
+
+To start the react app only:
+
+```yarn start```
+
+To start rails + the react app:
```
denys-davydov: Maybe we should add "bundle install" as a precondition to start rails server

client/.prettierrc.json
jonbwalker: Done! 80cee31

README.md
jonbwalker: I did as we discussed. eef5f5e

PR #34, opened 2018-05-25
CANS-186. Increased test coverage to 65% 

denys-davydov:
N.B. @jonbwalker is working on some other classes to cover with tests


client/src/components/Assessment/Domain.test.js
```diff
@@ -0,0 +1,45 @@
...
+  assessmentUnderSix={true}
+  i18n={clone(i18nPrototype)}
+  i18nAll={{'a': 'b'}}
+  onRatingUpdate={() => {}}
+  onConfidentialityUpdate={() => {}}
+/>;
+
+describe('<Domain />', () => {
+  it('renders', () => {
+    expect(() => shallow(domainComponentPrototype)).not.toThrow();
```
edgibbs: This render has multiple states, with different behavior.  This just addresses rendering without blowing up.

client/src/components/Assessment/I18nHelper.test.js
```diff
@@ -0,0 +1,35 @@
+import { clone } from 'lodash';
+import { getI18nByCode, getI18nValuesByPrefix } from './I18nHelper';
+
+describe('<I18nHelper />', () => {
+  describe('#getI18nByCode()', () => {
+    it('filters records by code and cuts the code off', () => {
```
edgibbs: Is code supposed to be local_code like en_US and that just isn't obvious from the test example?

client/src/components/Assessment/Item.test.js
```diff
@@ -0,0 +1,166 @@
...
+      />);
+      wrapper.setProps(clone(propsPrototype));
+      wrapper.instance().handleConfidentialityChange({target: {}})
+      expect(onConfidentialityUpdateMock.mock.calls.length).toBe(1);
+    });
+  });
+
+});
+
+const mountItem = item => {
```
edgibbs: This feels like it's floating at the bottom here, maybe within the describe where it's used or at least at the top of the test.

client/src/components/Assessment/Item.test.js
```diff
@@ -0,0 +1,166 @@
+import React from 'react';
+import { clone } from 'lodash';
```
edgibbs: Prefer Object.assign for cloning over loading lodash.

client/src/components/Assessment/Item.test.js
```diff
@@ -0,0 +1,166 @@
...
+      const item = clone(itemPrototype);
+      item.rating_type = "BOOLEAN";
+      const wrapper = mountItem(item);
+      wrapper.setProps(clone(propsPrototype));
+      const findSelect = wrapper.find(Select);
+      expect(findSelect.children().get(0).props.children.length).toBe(4);
+
+      wrapper.find(Icon).simulate('click');
+      const expandedText = wrapper.text();
+      expect(expandedText).toMatch(/No = rating 0 descriptionYes = rating 1 description/);
```
edgibbs: This regex is hard to follow as to what it's proving, do we actually display a 0 and a 1?

client/src/components/Assessment/Item.test.js
```diff
@@ -0,0 +1,166 @@
+import React from 'react';
+import { clone } from 'lodash';
+import { mount } from 'enzyme';
+import Item from './Item';
+import Icon from '@material-ui/core/Icon';
+import Radio from '@material-ui/core/Radio';
+import Select from '@material-ui/core/Select';
+
+const itemPrototype = {
```
edgibbs: Prototype is a pretty js code specific name, maybe just item, i18n, itemComponent.

client/src/setupTests.js
```diff
@@ -2,3 +2,20 @@ import { configure } from 'enzyme';
...
+Object.keys(document.defaultView).forEach((property) => {
+  if (typeof global[property] === 'undefined') {
+    exposedProperties.push(property);
+    global[property] = document.defaultView[property];
+  }
+});
+
+global.navigator = {
+  userAgent: 'node.js'
+};
```
edgibbs: no newline

client/src/setupTests.js
```diff
@@ -2,3 +2,20 @@ import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 configure({ adapter: new Adapter() });
+
+var jsdom = require('jsdom').jsdom;
```
edgibbs: OK, so this might be OK, but it rings off some alarm bells with testing anti-patterns.  I'm not sure why we're mocking globals like window, navigator, and document, seems like a YAGNI case.  We're also using 'var' instead of let or const.  Overall I'ld like to avoid mocking basic javascript components or at least mocking them much closer to where they're used.

client/src/components/Assessment/I18nHelper.test.js
denys-davydov: This is a code of a domain or an item inside of the assessment

client/src/components/Assessment/Domain.test.js
denys-davydov: Yes, it is.

client/src/components/Assessment/Item.test.js
denys-davydov: Moved it the file top where all the consts are initialized

client/src/components/Assessment/Item.test.js
denys-davydov: Using {...object} approach now

client/src/components/Assessment/Item.test.js
denys-davydov: Renamed the variables

client/src/components/Assessment/Item.test.js
denys-davydov: "No = rating 0 description" is a description for No value
"Yes = rating 1 description" is a description for Yes value

client/src/setupTests.js
denys-davydov: The tests used to fail without mocking window and document, so I've used the approach as a workaround. I've tried it now, and it works without the workaround now, so removing the mocks.

client/src/components/Assessment/I18nHelper.test.js
```diff
@@ -0,0 +1,35 @@
...
+    it('filters records by prefix and returns values only', () => {
+      const inputI18n = {
+        'item10._title_': 'Title 10',
+        'item10._description_': 'Description 10',
+        'item11._title_': 'Title 11',
+        'item11._description_': 'Description 11',
+      };
+      const actualResult = getI18nValuesByPrefix(inputI18n, 'item10.');
+      const expectedResult = ['Title 10', 'Description 10'];
+      expect(actualResult).toEqual(expectedResult)
```
jonbwalker: because of the filtering logic in getI18nValuesByPrefix, we should probably test a failure for this one too to make sure there's not a false positive

client/src/components/Assessment/Item.test.js
edgibbs: Weird but OK

PR #35, opened 2018-05-29
Updated ChildForm, added it to Page

abelayt:


client/src/components/ChildForm/ChildForm.js
```diff
@@ -0,0 +1,261 @@
...
+        last_name: '',
+        dob: '',
+        case_id: '',
+        county: {
+          id: 0,
+          name: '',
+        },
+      },
+    });
+    // window.location.reload();
```
edgibbs: remove comment.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -0,0 +1,261 @@
...
+  onFetchCountiesSuccess = counties => {
+    this.setState({
+      counties: counties,
+      counties_status: 'ready',
+    });
+  };
+
+  render() {
+    const { classes } = this.props;
+    // console.log('Child Info render: ', this.state.childInfo);
```
edgibbs: remove comment

client/src/components/ChildForm/ChildForm.service.js
```diff
@@ -0,0 +1,9 @@
+import appApi from '../../App.api';
+
+export class ChildFormService {
+  static createChild(child) {
+    return appApi.post(`/people`, child).then(response => response.data);
+  }
+}
+
+export default ChildFormService;
```
edgibbs: newline at end of file.

client/src/components/ChildForm/ChildForm.service.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const apiGetSpy = jest.spyOn(appApi, 'post');
+
+    beforeEach(() => {
+      apiGetSpy.mockReset();
+    });
+
+    it('posts ChildInfo Object', async () => {
+      const childInfo = {}
+      const expectedChildForm = `/people` ;
+      apiGetSpy.mockReturnValue(Promise.resolve({ data: `/people`}));
```
edgibbs: This probably returns something like a person object so maybe return something like { 'first_name': 'Anwar' }

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -0,0 +1,28 @@
...
+import TextField from '@material-ui/core/TextField';
+import Button from '@material-ui/core/Button';
+import ChildForm from './ChildForm';
+
+describe('<ChildForm />', () => {
+  const getWrapper = () => mount(<ChildForm  />);
+  const getLength = component => getWrapper().find(component).length;
+
+  it('renders without crashing', () => {
+    expect(() => shallow(<ChildForm />)).not.toThrow();
```
edgibbs: Since this successfully mounts on line 9 you can get away without this test.  It's a good first TDD test, but it's already proven at this point in the testing lifecycle.

client/src/components/ChildForm/Counties.service.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+describe('CountiesService', () => {
+  describe('#fetchCounties', () => {
+    const apiGetSpy = jest.spyOn(appApi, 'get');
+
+    beforeEach(() => {
+      apiGetSpy.mockReset();
+    });
+
+    it('returns array of Counties', async () => {
```
edgibbs: I like the async/await style the best out of the async testing options in jest, reads the cleanest, good use here.

client/src/components/ChildForm/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as ChildForm } from './ChildForm'
+export { default as CountiesService} from './Counties.service';
```
edgibbs: small newline needed here.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -0,0 +1,261 @@
...
+      </Fragment>
+    );
+  }
+}
+
+ChildForm.propTypes = {
+  classes: PropTypes.object.isRequired,
+};
+
+export default withStyles(styles)(ChildForm);
```
edgibbs: So this is using CSS in JS, style.  We probably need to convert back to SASS/CSS and just use classNames as this is Danny's direction with the CWDS Component library. I'm not sure given the timing when we want to deal with it.

abelayt: When we replace Material UI by CWDS/Component library, these will be gone with it, so no need to do it now.

client/src/components/ChildForm/ChildForm.js
abelayt: done

abelayt: done

client/src/components/ChildForm/ChildForm.js
abelayt: done

client/src/components/ChildForm/ChildForm.service.js
abelayt: done

client/src/components/ChildForm/index.js
abelayt: done

edgibbs:
The code climate issues are largely related to false positive dups since it still doesn't handle React very well.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -0,0 +1,261 @@
...
+  Card,
+  CardHeader,
+  CardContent,
+  CardActions,
+  Button,
+} from '@material-ui/core';
+
+import { CountiesService } from './Counties.service';
+import { ChildFormService } from './ChildForm.service';
+
```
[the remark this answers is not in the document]
abelayt: Material UI styling is different and  uses CSS-in-JS.  I had this discussion with you why doing  sass is not a priority for now.  Also why waste time doing sass on Material while the plan is to move to react-strap and most of it will change.

jonbwalker:
I'm slightly concerned about the state of the code here, being that it could go to production if we don't get the chance to swap out Material U.I. with @CWDS/components. I'll approve for now. but we should really start enforcing consistency in our project. It will result in less tech debt.

edgibbs:
Code climate will be mostly OK here once the config is setup.

PR #36, opened 2018-05-30
Cans 182 view child details

jonbwalker:


client/src/components/Client/Client.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  componentDidMount() {
+    this.fetchChildData(this.props.match.params.id);
+  }
+
+  fetchChildData(id) {
+    return PersonService.fetch(id)
+      .then(data => this.setState({ childData: data }))
+      .catch(() => console.log('No Child Data Found'));
```
edgibbs: I don't love the console log here, but I don't think I have a strong opinion either.

client/src/components/Client/person.service.js
```diff
@@ -0,0 +1,9 @@
+import appApi from '../../App.api';
+
+class PersonService {
```
edgibbs: This is more of a DDD concern but is client really Client or Child here, goes back to using Ubiquitious language.

client/src/components/Client/style.sass
```diff
@@ -0,0 +1,5 @@
+@import '../../styles/global/card.sass'
+
+.card-header
+  background-color: #114161
```
edgibbs: No way you'd know about this, but we do have some defined colors here, and I assume they've been carried over by Danny:

https://ca-cwds.github.io/react-wood-duck/?selectedKind=Styles&selectedStory=Colors&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

client/src/styles/global/card.sass
```diff
@@ -0,0 +1,4 @@
+@import "colors.sass"
+
+.card
+  min-width: 300px
```
edgibbs: We've also been using rems with 10px defined as 1 rem.

edgibbs: https://ca-cwds.github.io/react-wood-duck/?selectedKind=Styles&selectedStory=Typography&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

client/src/setupTests.js
```diff
@@ -2,3 +2,5 @@ import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 configure({ adapter: new Adapter() });
+
+global.XMLHttpRequest = undefined;
```
edgibbs: I think this is probably OK, for testing, but I'm not a big fan of global/singletons and resetting them.

jonbwalker: I can get rid of this, I thought this was needed for some testing, turns out it's not.

client/src/styles/global/card.sass
jonbwalker: updated! 06e1030

client/src/components/Client/Client.js
jonbwalker: I decided to go with a different approach and set childData to undefined on the state 2bf8371

edgibbs:
retest this please

edgibbs:
retest this please

edgibbs:
retest this please

client/src/components/Client/person.service.js
edgibbs: @david-sanabria clarified that it's supposed to be Person for the DB layer.

david-sanabria: Conceptually, `Client` is a relationship between a person and a case. The way I'd like to represent this within the data is that `client` is a dependent (child) entity, with a primary-foreign key to the `person` entity. The details specific to that relationship (e.g. `caseNumber`) should live in the `client` entity, and never in the `person` entity.

If we don't do it this way, we will end up with duplicate entries for people, to represent their various roles.

client/src/routes/Routes.js
```diff
@@ -1,14 +1,17 @@
...
   render() {
     return (
-      <Fragment>
-        <Route exact path="/" component={ClientsContainer}/>
-        <Route exact path="/assessments" component={Assessment}/>
-      </Fragment>
+      <Router>
+        <Fragment>
+          <Route path="/assessment" component={Assessment} />
+          <Route path="/client/:id" component={Client} />
```
denys-davydov: Restful approach is to use nouns in plural form for resources, so it is supposed to be "clients" if we are using the approach

client/src/components/Client/Client.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  componentDidMount() {
+    this.fetchChildData(this.props.match.params.id);
+  }
+
+  fetchChildData(id) {
+    return PersonService.fetch(id)
+      .then(data => this.setState({ childData: data }))
+      .catch(() => console.log(this.setState({ childData: undefined })));
```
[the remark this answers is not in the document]
edgibbs: +1 we're doing work in a console.log, seems bad.

PR #40, opened 2018-05-31
navigate to child detail page on successful new child creation

jonbwalker:


client/src/components/ChildForm/ChildForm.js
```diff
@@ -144,10 +146,14 @@ class ChildForm extends Component {
   };
 
   render() {
+    if (this.state.child_status === 'ready') {
+      return <Redirect to={'/clients/' + this.state.childInfo.id} />;
```
edgibbs: test?

denys-davydov:
@edgibbs @jonbwalker So the root cause of the issue is that we are using a Material UI components library and a styles wrapper component for it. The wrapper wraps the ChildForm component and enzyme cannot test it in a usual manner, because enzyme is supposed to work with the root component, and the root component in this case is WithStyles.

PR #41, opened 2018-06-03
Creates a new feature spec to test clients

edgibbs:
## Description

Handles testing of client pages and also handles perry session logins across tests.  May want to change to logout down the road, but reasonable for now.

## Tests

## Types of changes

## Checklist:


spec/acceptance/client_page_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    fill_in('First Name', with: '')
+    click_button('Save')
+    expect(page).to have_content 'Add Child/Youth'
+
+    fill_in('First Name', with: first_name)
+    fill_in('Last Name', with: last_name)
+    page.find('#dob').set(date_of_birth)
+    fill_in('Case Number', with: case_number)
+    find('.jss96').click
```
denys-davydov: This '96' suffix looks like a generated number, so any change on the page will lead to the number to be changed and test to become broken. Can we use a different selector if it really works that way?

edgibbs: Good catch, I'll take another quick pass at it, the issue is related to how the component is rendered and at some point I just moved on, but there should be some way to put a better marker on it.

edgibbs: OK, updated to not rely on a generated class name.

spec/acceptance/client_page_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
jonbwalker: Does something use frozen_string_literal? I've seen this in some of our other files.

edgibbs: Oh, it's a future version of ruby thing for performance enhancements:'

https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html

PR #42, opened 2018-06-04
New Child form Validations CANS 188

abelayt:
## Tests

## Types of changes

## Checklist:


client/src/components/ChildForm/ChildForm.js
```diff
@@ -108,6 +116,8 @@ class ChildForm extends Component {
         });
       })
       .catch(() => this.setState({ child_status: 'error' }));
+
+    window.alert('Form Submitted Successfully!');
```
denys-davydov: Don't forget to remove the line

docker-compose.yml
```diff
@@ -1,68 +1,28 @@
-version: '3.2'
+version: '2.1'
```
denys-davydov: I suppose the file changes are committed by mistake

abelayt: Will fix that.

client/src/components/ChildForm/ChildForm.js
abelayt: This is for notification of successful submission of the form and one of the acceptance criteria.

docker-compose.yml
```diff
@@ -1,68 +1,28 @@
...
-    image: cwds/perry:3.0.1_581-RC
-    hostname: perry
-    ports:
-      - ${PERRY_PORT}:8080
-      - ${PERRY_ADMIN_PORT}:9082
-    environment:
-      PERRY_PORT: ${PERRY_PORT}
-      DEV_MODE: ${DEV_MODE}
-      LOG_LEVEL: ${LOG_LEVEL}
+    env_file: .env
```
edgibbs: I'd swap it back to master, but the .env file reference is breaking the build.

client/src/components/ChildForm/ChildForm.helper.js
```diff
@@ -0,0 +1,26 @@
+export function validate(fieldName, value) {
+  switch (fieldName) {
```
edgibbs: A bit of cyclomatic complexity here, but it reads well and has good test coverage so I think it's fine.

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,142 @@
...
+      case_id: '',
+      county: {
+        id: 0,
+        name: '',
+      },
+    });
+  });
+
+  it('#handleSubmit()', async () => {
+    const wrapper = await getWrapper();
```
edgibbs: This test doesn't have an assertion/expectation.

docker-compose.yml
abelayt: fixed.

client/src/components/ChildForm/ChildForm.js
jonbwalker: Use of alert is not ideal and should be avoided mainly because the user can't proceed until they click 'Ok" to dismiss the alert window. We should use different form of notification to display the success message.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -145,15 +156,27 @@ class ChildForm extends Component {
...
     if (this.state.child_status === 'ready') {
-      return <Redirect to={'/clients/' + this.state.childInfo.id} />;
+      return <Redirect to={`/clients/${this.state.childInfo.id}`} />;
     }
-
     const { classes } = this.props;
     return (
       <Fragment>
-        <PageInfo title={'Add Child/Youth'} />
+        <h3>Add Child/Youth</h3>
```
jonbwalker: Was PageInfo removed by accident? It should be here.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -238,14 +270,15 @@ class ChildForm extends Component {
             </form>
           </CardContent>
           <CardActions>
-            <Button onClick={this.handleCancel} href="/">
+            <a onClick={this.handleCancel} href="/clients/new">
```
jonbwalker: We should use the React Router <Link /> component for this href so we don't force a full page reload.

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,142 @@
 import React from 'react';
-import {shallow, mount } from 'enzyme';
-import Card from '@material-ui/core/Card';
-import TextField from '@material-ui/core/TextField';
-import Button from '@material-ui/core/Button';
+import { mount, shallow } from 'enzyme';
```
jonbwalker: mount, shallow and render are all in the setupTests.js so they don't need to be imported into each individual test file anymore.

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,142 @@
...
+        true
+      );
+      expect(childFormWrapper.state('isSaveButtonDisabled')).toEqual(true);
+    });
+
+    it('enables Save button when all validations are good', async () => {
+      const wrapper = await getWrapper();
+      const childFormWrapper = wrapper.find('ChildForm').dive();
+      const childFormInstance = childFormWrapper.instance();
+      await childFormInstance.handleChange('first_name')({
```
jonbwalker: I know the below 'John's' are all just test values but it makes the readability a little strange to have 'John' as case_id and dob.

client/src/components/ChildForm/ChildForm.test.js
denys-davydov: You still need these imports if you are running a test separately

jonbwalker: Can you elaborate on this? It's my understanding that jest will pull in setupTests.js every time it runs.

client/src/components/ChildForm/ChildForm.js
abelayt: Yes, but do you have any suggestion? Also this is as part of temporary solution not to do it twice since this and all will be changed to cwds/components, right?

client/src/components/ChildForm/ChildForm.js
abelayt: Yes it was and now restored, thanks for the catch!
Note: I'm not sure if importing and inserting a component to just style a title is and efficient way. But this and related way of doing styles is subject to discussion.

client/src/components/ChildForm/ChildForm.js
jonbwalker: I would use Material UI Snackbar for now (see link below).

https://material-ui.com/demos/snackbars/

Another option would be to show/hide a div with a success message using component state on successful submit:
https://stackoverflow.com/questions/24502898/show-or-hide-element-in-react

Given our timeline, there's a good chance that the code we're writing right now will end up in production, so we shouldn't introduce any unideal UX workflows on the terms of thinking that it's temporary because it's likely that we won't have the time to fix it before our July deadline.

client/src/components/ChildForm/ChildForm.js
jonbwalker: The reason I took this approach is so we have one source of truth for our page titles, that way if we need to change anything (col spacing, margin, padding, width, height, etc...) then we only have to do it in once place and it will be reflected on all of our pages. It's one of the main benefits of components! ;)

client/src/components/ChildForm/ChildForm.js
abelayt: I tried that, can you please slack me the snippet code to do that. Thanks!
Note: The React Router is not working properly. On click, the path changes but needs a refresh to load the page. That is the case on my side.

jonbwalker: I just pushed a PR that will fix the router Link issue: 
https://github.com/ca-cwds/cans/pull/45

please review!

client/src/components/ChildForm/ChildForm.js
abelayt: Of course that is the obvious and no confusion on that. But we are talking about a CardHeader, and is part of a Card. It is better to style it directly with  global style than another component like PageInfo that just implements a 'title style'. Even the naming is misleading. I think we need to have discussion about these and for now PageInfo is restored.

jonbwalker: It wasn't just imported for 'title' styles, it brings in Row, Col from reactstrap so it's situated on the page correctly and consistently across all pages so it doesn't need to be repositioned  on each page, again, one of the benefits of using components.

If we had only needed the 'title' style, then I would have taken the approach a global css style to use across all pages.

client/src/components/ChildForm/ChildForm.js
jonbwalker: The PR mentioned above was merged into master, update your branch with master and you can use '<Link />' here

abelayt: 1. I tried the link and getting the test failed with : "Invariant Violation: You should not use <Link> outside a <Router>"
2. Using the Link does not clear the form since no reload. I want to clear/reset the form on Cancel. Reverting to href till better solution found.

client/src/components/ChildForm/ChildForm.test.js
denys-davydov: When I start the test in Intellij IDEA for debugging it requires the implicit import for render, shallow and mount. So I think it's OK to have these imports for the convenience of development.

edgibbs: As long as we're not requiring unused items in the file I guess I'm OK with explicit declarations.  I'm used to explicit dependencies on things like setupJs

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,134 @@
...
+      person_role: 'CLIENT',
+      first_name: '',
+      last_name: '',
+      dob: '',
+      case_id: '',
+      county: {
+        id: 0,
+        name: '',
+      },
+    });
```
edgibbs: #handleSubmit() is never tested.

client/src/components/ChildForm/ChildForm.js
jonbwalker: 1. You have to mock React Router using Memory Router. In your test, wrap your mount/shallow call in MemoryRouter:
```shallow(<MemoryRouter><Childform ><MemoryRouter>)```
2. You can programmatically clear the form by resetting the object on successful submit: 
`` this.setState({ childInfo: undefined });```

File: client/src/components/ChildForm/ChildForm.test.js (hunk 8 above)
Comment (jonbwalker) [reply client/src/components/ChildForm/ChildForm.test.js]: @denys-davydov Okay, that's interesting. Should we just remove the global import from setupTests.js then and include the imports in all of the files for sake of consistency?

File: client/src/components/ChildForm/ChildForm.test.js (hunk 10 above)
Comment (abelayt) [reply client/src/components/ChildForm/ChildForm.test.js]: Working on preventDefault to get a passing test with assert.

File: client/src/components/ChildForm/ChildForm.test.js (hunk 8 above)
Comment (edgibbs) [reply client/src/components/ChildForm/ChildForm.test.js]: Yeah it's a test so I think DRY is even less applicable here.

File: client/src/components/ChildForm/ChildForm.test.js (hunk 10 above)
Comment (abelayt) [reply client/src/components/ChildForm/ChildForm.test.js]: Done.

File: client/src/components/ChildForm/Notification.js [hunk 11]
```diff
@@ -0,0 +1,128 @@
...
+      className={classNames(classes[variant], className)}
+      aria-describedby="client-snackbar"
+      message={
+        <span id="client-snackbar" className={classes.message}>
+          <Icon className={classNames(classes.icon, classes.iconVariant)} />
+          {message}
+        </span>
+      }
+      action={[
+        <IconButton
```
jonbwalker: Will the user ever need to manually close the success message?

abelayt: Yes by clicking on X, no need to wait for the timer.

jonbwalker: Ok, that makes sense.

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -103,6 +102,15 @@ describe('<ChildForm />', () => {
...
 
+  it('#handleSubmit()', async () => {
+    const fakeEvent = { preventDefault: () => {} };
+    const wrapper = await getWrapper();
+    const childFormWrapper = wrapper.find('ChildForm').dive();
+    expect(childFormWrapper.find('form').length).toBe(1);
+    expect(await getLength('TextField')).toEqual(5);
+    childFormWrapper.find('form').simulate('submit', fakeEvent);
+  });
+
```
jonbwalker: Do we need to test what happens after this line: 
``` childFormWrapper.find('form').simulate('submit', fakeEvent);```

File: client/src/components/ChildForm/ChildForm.js [hunk 13]
```diff
@@ -72,20 +73,31 @@ class ChildForm extends Component {
...
+        last_name: false,
+        dob: false,
+        case_id: false,
+        county: false,
+      },
+      isSaveButtonDisabled: true,
       open: false,
+      navigate: false,
     };
     this.handleChange = this.handleChange.bind(this);
```
jonbwalker: Binding shouldn't be needed when you use arrow functions.
Did you experience something that caused you need to bind these functions?

client/src/components/ChildForm/ChildForm.test.js
denys-davydov: It is nice to the global imports, and I have no issues with having local imports in some places too.

client/src/components/ChildForm/ChildForm.js
```diff
@@ -238,21 +272,22 @@ class ChildForm extends Component {
...
               color="primary"
+              disabled={this.state.isSaveButtonDisabled}
               className={classes.button}
               onClick={this.handleSubmit}
             >
               Save
             </Button>
           </CardActions>
         </Card>
+        {this.state.child_status === 'ready' && <Notification />}
```
edgibbs: should this use the new `navigate` property like line 174

client/src/components/ChildForm/Notification.js
```diff
@@ -0,0 +1,130 @@
...
+  SnackbarContent,
+} from '@material-ui/core';
+
+const variantIcon = {
+  success: CheckCircleIcon,
+  backgroundColor: green,
+};
+
+const styles1 = theme => ({
+
```
edgibbs: looks like a bonus newline

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,141 @@
...
-    expect(getLength(Button)).toBe(2);
+  it('renders counties drop down', async () => {
+    const wrapper = await getWrapper();
+    const childFormWrapper = wrapper.find('ChildForm').dive();
+    const childFormInstance = childFormWrapper.instance();
+    childFormInstance.onFetchCountiesSuccess([{ id: '1', name: 'Alameda' }]);
+    childFormWrapper.update();
+    const menuItems = childFormWrapper.find('WithStyles(MenuItem)');
+    expect(menuItems.length).toEqual(2);
+    expect(menuItems.get(1).props.children).toEqual('Alameda');
```
edgibbs: This is wrapped in a span now, so the test is failing.

client/src/components/ChildForm/Notification.js
```diff
@@ -0,0 +1,130 @@
...
+MessageContent.propTypes = {
+  classes: PropTypes.object.isRequired,
+  className: PropTypes.string,
+  message: PropTypes.node,
+  onClose: PropTypes.func,
+  variant: PropTypes.oneOf(['success']).isRequired,
+};
+const MessageContentWrapper = withStyles(styles1)(MessageContent);
+const styles2 = theme => ({
```
edgibbs: Trying to think of a more descriptive name then styles1 or styles2

client/src/components/ChildForm/Notification.js
```diff
@@ -0,0 +1,130 @@
...
+    opacity: 0.9,
+    marginRight: theme.spacing.unit,
+  },
+  message: {
+    display: 'flex',
+    alignItems: 'center',
+  },
+});
+
+let MessageContent = (props) => {
```
edgibbs: I think this can just be a const, we're not reassigning it.

client/src/components/ChildForm/Notification.js
edgibbs: It's a little weird at the bottom of the card, but seems fine for now since Ken is going to figure out the flow for a future story.

client/src/components/ChildForm/Notification.test.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { mount, shallow } from 'enzyme';
+import Notification from './Notification';
+
+describe('<Notification />', () => {
+  const getWrapper = async () => await shallow(<Notification />);
```
edgibbs: I think this async/await isn't needed for a shallow render.

client/src/components/ChildForm/Notification.test.js
```diff
@@ -0,0 +1,27 @@
...
+  const getMountWrapper = async () => await mount(<Notification />);
+  const getLength = async component => {
+    const wrapper = await getMountWrapper();
+    const length = wrapper.find(component).length;
+    wrapper.unmount();
+    return length;
+  };
+
+  it('verifies the numbers of SnackbarContent', async () => {
```
edgibbs: There are so many async awaits here it's hard to follow.  I assuming this was the combination that passed.  I think I'd just move almost all the of the setup outside the tests into the tests.

client/src/components/Layout/Page.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { Container, Row, Col } from 'reactstrap';
 import { PageInfo, SideNav } from './';
 import { Routes } from '../../routes';
-import { ChildForm } from '../ChildForm';
+
```
edgibbs: This seems like an extra line.

client/src/components/ChildForm/ChildForm.js
abelayt: First when the  child_status === 'ready' then navigate: true and redirect is setTimeout to go after 5 sec at line 118. Also when child_status === 'ready'  <Notification /> message displays for 5 sec and by then  the setTimeout expires and redirect is done.

client/src/components/ChildForm/ChildForm.test.js
abelayt: I don't see any test failing.

client/src/components/ChildForm/Notification.js
abelayt: Yes, changed to const

client/src/components/ChildForm/ChildForm.test.js
```diff
@@ -1,24 +1,141 @@
...
+        target: { value: 'John' },
+      });
+      await childFormInstance.handleChange('county')({
+        target: { value: { id: '1' } },
+      });
+      expect(childFormWrapper.state('isSaveButtonDisabled')).toEqual(false);
+    });
+  });
+
+  it('#handleSubmit()', async () => {
```
jonbwalker: There's quite a bit of stuff going on in the handleSubmit function, for example it set's state twice. I think we need a bit more coverage of the logic happening there.

client/src/components/Layout/Page.js
abelayt: Needed this to get a link to the add child form on the sidebar.

client/src/components/ChildForm/Notification.test.js
abelayt: Okay, removed.

client/src/components/ChildForm/ChildForm.js
edgibbs: OK, makes sense.

client/src/components/ChildForm/ChildForm.test.js
edgibbs: Cool.

client/src/components/ChildForm/Notification.test.js
edgibbs: OK, it's passing, can look at a refactor later.

edgibbs:
Alright there's an acceptance test that needs to be updated, but it was added after this story was underway, so there will be a small PR to fix it.  Codeclimate issues are all false positives on dups or a few line limits.

PR #52, opened 2018-06-12
Adding intial code climate

edgibbs:


.codeclimate.yml
```diff
@@ -0,0 +1,43 @@
...
+  method-lines:
+    enabled: true
+    config:
+      threshold: 50
+  nested-control-flow:
+    enabled: true
+  return-statements:
+    enabled: true
+  similar-code:
+    enabled: false
```
denys-davydov: 👍

.codeclimate.yml
```diff
@@ -0,0 +1,43 @@
...
+plugins:
+  brakeman:
+    enabled: true
+  rubocop:
+    enabled: true
+exclude_patterns:
+  - "config/"
+  - "db/"
+  - "dist/"
+  - "features/"
```
denys-davydov: We don't have db/, dist/ and features/ folders in our repo. Do we really need to have them in exclude_patterns?

edgibbs: Good catch 🐟.  We have a config/, but I'll delete the others

PR #65, opened 2018-06-25
Patches sprockets gem to fix CVE-2018-3760

edgibbs:


edgibbs:
Looks like another pull request must have already updated sprockets, so closing.

PR #70, opened 2018-07-01
CANS-202 Adding eslint/prettier checks

edgibbs:
Lots of small changes to fix linting, prettier, and a few test fixes.  The only item I was unable to fix completely was:

app/javascript/Application/components/Layout/Page.test.js

Had to pend one of the specs for now.

https://osi-cwds.atlassian.net/browse/CANS-238


app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -254,23 +221,22 @@ describe('<Assessment />', () => {
         const wrapper = shallow(<Assessment {...props} />);
 
         wrapper.instance().componentDidMount();
-        expect(assessmentServiceGetSpy).toHaveBeenCalled();
+        expect(assessmentServiceGetSpy).toHaveBeenCalledWith();
```
jonbwalker: I believe `.toHaveBeenCalledWith()` needs an argument here.

edgibbs: Yeah, it looked a little weird to me as well, but that is apparently the way to test for no-args in jest:

https://github.com/facebook/jest/issues/1899

I'd like something like `toHaveBeenCalledWith(noArgs)` but that's not how it works, maybe I'll work up a PR on jest.

jonbwalker: Okay, that makes sense, thanks for the source. I figured there was a good reason for the change but it looked a little strange.

PR #75, opened 2018-07-05
fix Page tests

jonbwalker:
* swap mount with shallow
* remove react-router MemoryRouter (not needed)
* add additional describe layer

## Tests

## Types of changes

## Checklist:


jonbwalker:
@edgibbs The issue was with the way SideNav was being imported, for some reason, enzyme wasn't recognizing it as a component. So fixing the import is what actually fixed the test.

The reason I swapped out mount for shallow is the tests we have don't need a fully mounted component (the component being mounted, plus a full render of all child components and their children, plus some additional react lifecycle hooks that we aren't leveraging). We can simply do a shallow render and then dive down into one of the child components to verify that it's doing what we expect it to. It's essentially just a spot check.

PR #76, opened 2018-07-09
CANS-236 + CANS-248: Add/Remove caregiver domains

denys-davydov:
## Description
https://osi-cwds.atlassian.net/browse/CANS-236
https://osi-cwds.atlassian.net/browse/CANS-248

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -1,30 +1,78 @@
...
 import { defaultEmptyAssessment } from './AssessmentHelper';
 import { getI18nByCode } from './I18nHelper';
 import Domain from './Domain';
 
+const INDICES = 'abcdefghijklmnopqrstuvwxyz';
+
 class Assessment extends Component {
-  handleUpdateItemRating = (code, rating) => {
-    this.updateItem(code, 'rating', rating);
+  /* eslint-disable camelcase */
```
jonbwalker: The react docs advise to avoid using `componentWillReceiveProps` in new code because it can lead to bugs and inconsistencies. They suggest using a `memorization helper` instead.

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -70,9 +59,46 @@ class Domain extends Component {
...
+    this.props.onRemoveCaregiverDomain(this.props.domain.caregiver_index);
+  };
+
+  renderCaregiverDomainControls() {
+    return (
+      <div className={'caregiver-domain-controls'}>
+        <h5>
+          <div
+            onClick={this.handleRemoveCaregiverDomain}
+            onKeyPress={this.handleRemoveCaregiverDomain}
```
jonbwalker: This appears to be on any key press, we should probably isolate it to just when just the delete key is pressed.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -70,9 +59,46 @@ class Domain extends Component {
...
+            role={'button'}
+            tabIndex={0}
+          >
+            - REMOVE CAREGIVER
+          </div>
+        </h5>
+        <h5>
+          <div
+            onClick={this.handleAddCaregiverDomain}
+            onKeyPress={this.handleAddCaregiverDomain}
```
jonbwalker: same as above but enter key.

app/javascript/Application/components/Assessment/Domain.js
denys-davydov: It is an a11y issue, not sure if it should be a delete key or any other particular key. So I have any key pressed when the element is focused.

app/javascript/Application/components/Assessment/Domain.js
denys-davydov: The same as above

app/javascript/Application/components/Assessment/Assessment.js
denys-davydov: We have 3 places in the code where we use _componentWillReceiveProps()_ function now, so I believe it would be better to fix them together in a separate PR as a part of code cleaning.

app/javascript/Application/components/Assessment/Domain.js
jonbwalker: As a user, I would not expect this behavior. And it could potentially be confusing for some users.

denys-davydov: I agree it is confusing that any key activates an action, as it is still confusing for the delete key. So I've found this document https://webaim.org/techniques/keyboard/ about web accessibility and it says that for links Enter key should be used.

jonbwalker: Actually yes, the enter key does make more sense.

denys-davydov:
retest this please

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -15,25 +17,102 @@ describe('<Assessment />', () => {
...
+      // when
+      instance.addCaregiverDomainAfter('a');
+
+      // then
+      const newAsmnt = mockFn.mock.calls[0][0];
+      expect(newAsmnt.state.domains.map(domain => domain.caregiver_index)).toEqual(['a', 'b']);
+    });
+
+    it('removes caregiver domain', () => {
+      // given
```
edgibbs: This test is hard to follow, some of that is having the assessment defined in another file, but a lot of it is that it looks like the test has three domains in the setup with a,b,c for the caregiver_indexes then you remove 'b' and the expectation is that it was called with ['a', 'b'] which is confusing, it looks like it removed c.

edgibbs: After spending about 5 minutes messing with it and console logging some I realize it removes the last domain and renames them so that the resulting set is ['a', 'b'].  Maybe a description like 'removes the caregiver domain and resets the alpha names of the remaining domain caregiver indexes'  (OK, that seems a bit long)

app/javascript/Application/components/Assessment/Domain.js
edgibbs: +1

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -15,25 +17,102 @@ describe('<Assessment />', () => {
...
 
-      it('is invoked when item confidentiality is updated', () => {
-        const mockFn = jest.fn();
-        const wrapper = mount(<Assessment onAssessmentUpdate={mockFn} assessment={assessment} i18n={i18n} />);
-        expect(assessment.state.domains[0].items[3].confidential).toBe(false);
-        wrapper.instance().handleUpdateItemConfidentiality('EXPOSURE', true);
-        const newAssessment = cloneDeep(assessment);
-        newAssessment.state.domains[0].items[3].confidential = true;
-        expect(mockFn).toHaveBeenCalledWith(newAssessment);
+      describe('confidentiality', () => {
```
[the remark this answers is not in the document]
denys-davydov: Yep, you got the function name two describes above

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -1,30 +1,78 @@
...
+    assessmentState.domains
+      .filter(domain => domain.is_caregiver_domain)
+      .map(domain => (domain.caregiver_index = INDICES[i++]));
+    return assessmentState;
+  }
+
+  addCaregiverDomainAfter = caregiverIndex => {
+    const assessment = cloneDeep(this.props.assessment);
+    const domains = assessment.state.domains;
+    for (let i = 0; i < domains.length; i++) {
```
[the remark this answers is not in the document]
denys-davydov: Not sure it will work with const index, I'll prefer to have the old style time-tested approach for now :)

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -15,25 +17,102 @@ describe('<Assessment />', () => {
...
+          wrapper.instance().handleUpdateItemRating('1', 2);
+          const newAssessment = cloneDeep(assessment);
+          newAssessment.state.domains[0].items[0].rating = 2;
+          expect(mockFn).toHaveBeenCalledWith(newAssessment);
+        });
+
+        it('is invoked when item rating is updated for caregiver domain', () => {
+          // given
+          const mockFn = jest.fn();
+          const asmnt = cloneDeep(assessment);
```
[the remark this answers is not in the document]
denys-davydov: Renamed all over the class

app/javascript/Application/components/Assessment/Assessment.test.js
denys-davydov: Renamed to make it more descriptive

app/javascript/Application/components/Assessment/Assessment.js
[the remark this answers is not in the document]
edgibbs: you apparently don't hate for loops with `i` varaibles as much as I do.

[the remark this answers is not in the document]
denys-davydov: Java background makes more conservative

PR #129, opened 2018-08-31
CANS-101. Print assessment

denys-davydov:
## Description
https://osi-cwds.atlassian.net/browse/CANS-101

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Print/PrintAssessment.js
```diff
@@ -0,0 +1,172 @@
...
+
+  render() {
+    const { isAssessmentUnderSix } = this.state;
+    const { i18n } = this.props;
+    const domains = this.props.assessment.state.domains;
+    return (
+      <div>
+        {this.renderHeader()}
+        {domains.map(domain => {
+          if (!shouldDomainBeRendered(isAssessmentUnderSix, domain)) return null;
```
edgibbs: So this gives a lot of fine grained control, but it is also tightly coupled to changes in assessments as in when assessments change this will probably have to change with similar logic.

app/javascript/Application/components/Print/PrintAssessment.test.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import PrintAssessment from './PrintAssessment';
+import { assessmentPrint, i18nPrint, printFixture } from '../Assessment/assessment.mocks.test';
+
+describe('<PrintAssessment />', () => {
+  it('should match fixture as plain html', () => {
```
edgibbs: This test seems like it's pretty brittle with any changes to assessments, my thought would be to unit test the logic in the PrintAssessment component like domains that should not be rendered if under_six.

app/javascript/Application/components/Print/Print.js
```diff
@@ -0,0 +1,34 @@
...
+    frame.focus();
+    frame.document.open();
+    frame.document.write(renderToString(this.props.node));
+    frame.document.close();
+    frame.print();
+    this.props.onClose();
+  }
+
+  render() {
+    return <iframe id={printFrameId} title={printFrameId} />;
```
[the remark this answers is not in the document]
edgibbs: Talked with Denys this was arrived at after having issues with competing stylesheets and components, reasonable solution.

app/javascript/Application/components/Print/PrintAssessment.test.js
edgibbs: Had a quick discussion and going with this approach for now.

edgibbs:
Code Climate is just complaining about a large fixture in a test, doesn't need to be resolved.

PR #136, opened 2018-09-05
CANS-291: AssessmentContainer tests fix

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


edgibbs:
They actually weren't running because the async on the describe wasn't syntactically wrong, but it doesn't work with jest so the tests were never running.  Moving the async block down to the `it` level allows them to run.

PR #165, opened 2018-09-17
CANS-285 Prevent Unauthorized Users from Accessing CANS

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-285

## Tests

- [x ] I used TDD to develop the feature

## Types of changes

## Checklist:


spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  describe CwdsPermissionChecker do
+    describe '#call' do
+      let(:application) { instance_double('ActionDispatch::Routing::RouteSet') }
+      let(:cwds_permission_checker) { CwdsPermissionChecker.new(application) }
+
+      context 'when there is no session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+        let(:request) { Rack::Request.new(environment) }
```
edgibbs: Small thing, but there's usually a line break after the last let block and before the it.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+          expect(status).to eq 301
+          expect(headers['Location']).to eq '/error_page'
+        end
+      end
+
+      context 'when there is invalid session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+        let(:request) { Rack::Request.new(environment) }
```
edgibbs: linebreak

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,75 @@
...
```
edgibbs: linebreak

lib/infrastructure/cwds_permission_checker.rb
```diff
@@ -0,0 +1,32 @@
...
+    def call(environment)
+      request = Rack::Request.new(environment)
+      return redirect_to_error_page(request.url) unless valid_roles_privileges(request)
+      @application.call(environment)
+    end
+
+    private
+
+    def valid_roles_privileges(request)
+      request.session['roles']&.include?('CANS-worker') &&
```
edgibbs: Feels like this needs a spec covering when you don't get back any roles in a session for some reason.

nasar-cambria:
@edgibbs - Thanks, your suggestions really helped. Added one more missing scenario which was when there is a valid session the user should be able to access application.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
nasar-cambria: fixed

lib/infrastructure/cwds_permission_checker.rb
nasar-cambria: Have a test case covering this scenario on line 61.

nasar-cambria:
re-test this please

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      let(:application) { instance_double('ActionDispatch::Routing::RouteSet') }
+      let(:cwds_permission_checker) { CwdsPermissionChecker.new(application) }
+
+      context 'when there is no session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+
+        it 'redirects when user has no roles and privileges' do
+          allow(cwds_permission_checker).to receive(:valid_roles_privileges)
```
edgibbs: Ouch, missed this, you're stubbing out the class under test here so you're not actually testing the `valid_roles_and_privileges()` method.  These tests all need to be reworked.  You can comment out the body of the method and everything still passes.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          expect(headers['Location']).to eq '/error_page'
+        end
+      end
+
+      context 'when there is a valid session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+
+        it 'user is able to access the application' do
```
edgibbs: Same comment here.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+
+      context 'when there is invalid session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+        let(:request) { Rack::Request.new(environment) }
+
+        it 'redirects when user has no privileges' do
+          allow(application).to receive(:call)
```
edgibbs: This line is never called and shouldn't be stubbed.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+
+      context 'when there is invalid session' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+        let(:request) { Rack::Request.new(environment) }
+
+        it 'redirects when user has no roles' do
+          allow(application).to receive(:call)
```
edgibbs: Again never called shouldn't be stubbed.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      context 'when session return valid roles and privileges' do
+        let(:environment) do
+          Rack::MockRequest.env_for('http://example.com?accessCode=good_accessCode', {})
+        end
+        let(:valid_roles_privileges) do
+          request.session['roles'] = 'CANS-worker' && request.session['privileges'] = 'CANS-rollout'
+        end
+
+        it 'returns valid roles from session' do
```
edgibbs: The assertion here should be that it calls `@application.call(environment)`  The session is never changed so asserting that it hasn't changed doesn't prove anything.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          request = Rack::Request.new(environment)
+          request.session['roles'] = 'CANS-worker'
+          _status, _headers, = cwds_permission_checker.call(environment)
```
edgibbs: You don't need to have _status or _headers it's not required by ruby.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          expect(request.session['roles']).to eq 'CANS-worker'
+        end
+        it 'returns valid privileges from session' do
+          request = Rack::Request.new(environment)
+          request.session['privileges'] = 'CANS-rollout'
+          _status, _headers, = cwds_permission_checker.call(environment)
```
edgibbs: _status and _headers don't need to exist.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        let(:valid_roles_privileges) do
+          request.session['roles'] = 'CANS-worker' && request.session['privileges'] = 'CANS-rollout'
+        end
+
+        it 'returns valid roles from session' do
+          request = Rack::Request.new(environment)
+          request.session['roles'] = 'CANS-worker'
+          _status, _headers, = cwds_permission_checker.call(environment)
+          expect(request.session['roles']).to eq 'CANS-worker'
+        end
```
edgibbs: Needs a newline after it.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        it 'returns valid privileges from session' do
+          request = Rack::Request.new(environment)
+          request.session['privileges'] = 'CANS-rollout'
+          _status, _headers, = cwds_permission_checker.call(environment)
+          expect(request.session['privileges']).to eq 'CANS-rollout'
```
edgibbs: Again this doesn't assert anything of value since the session state is never changed by any of the code paths.

lib/infrastructure/cwds_permission_checker.rb
edgibbs: There's a few issues with that spec class.

spec/lib/infrastructure/cwds_permission_checker_spec.rb
nasar-cambria: This is fixed now, I have updated test case to stub session with valid roles and privileges to make sure it fails when there is no roles or privileges.

nasar-cambria:
@edgibbs re worked on all test cases . Can you please re visit when you get a chance?

spec/lib/infrastructure/cwds_permission_checker_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      context 'when there is a valid session' do
+        let(:request) do
+          request = Rack::Request.new(environment)
+          request.session['roles'] = ['CANS-worker']
+          request.session['privileges'] = ['CANS-rollout']
+          request
+        end
+
+        it 'user is able to access the application' do
+          result = cwds_permission_checker.send('valid_roles_privileges', request)
```
edgibbs: OK we're testing a private method here, Ruby is cool like that, but just because you can.  Anyway you're testing through the public interface which is:

```ruby
CwdsPermissionChecker#call(environment)
```
You should be able to test everything because there are only 2 possible exits:

- You test that a redirect happened
- Or you assert that it returns a 200 status

edgibbs: The other option here if you want to test just the roles/sessions is to move the logic to another small class and then unit test it there and just stub the return here, maybe something like:

```ruby
RolesAndPermissionsValidator
```

There's probably a better name here.

PR #173, opened 2018-09-19
CANS-285 Update url Path

nasar-cambria:
## Description
Updating url path to cans/error_page. cans_base_path is not creating cans in the url so updating it inside the method. This should work in pre int and int.


lib/infrastructure/cwds_permission_checker.rb
```diff
@@ -26,8 +26,7 @@ def redirect_to_error_page(_url)
     end
 
     def error_page_url
-      cans_base_path = Rails.configuration.micro_services['cans_base_path']
-      "#{cans_base_path}/error_page"
+      'cans/error_page'
```
edgibbs: The hard coding here is dangerous because they may change this in the future, better to pull from config.

edgibbs: Just tested it locally with /cans enabled, I get the following in my URL when redirected:

http://localhost:3000/cans/cans/error_page

I'm not sure we need to mention /cans at all since rails appears to be handling it fine.

public/cans/error_page.html
```diff
@@ -3,8 +3,10 @@
   <head>
     <title>The change you wanted was rejected (422)</title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.3.1/css/all.css">
```
edgibbs: Don't love hardcoded style link.

lib/infrastructure/cwds_permission_checker.rb
nasar-cambria: Updated as discussed.

public/exclamation-triangle-solid.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" data-prefix="fas" data-icon="exclamation-triangle" class="svg-inline--fa fa-exclamation-triangle fa-w-18" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 576 512"><path fill="currentColor" d="M569.517 440.013C587.975 472.007 564.806 512 527.94 512H48.054c-36.937 0-59.999-40.055-41.577-71.987L246.423 23.985c18.467-32.009 64.72-31.951 83.154 0l239.94 416.028zM288 354c-25.405 0-46 20.595-46 46s20.595 46 46 46 46-20.595 46-46-20.595-46-46-46zm-43.673-165.346l7.418 136c.347 6.364 5.609 11.346 11.982 11.346h48.546c6.373 0 11.635-4.982 11.982-11.346l7.418-136c.375-6.874-5.098-12.654-11.982-12.654h-63.383c-6.884 0-12.356 5.78-11.981 12.654z"></path></svg>
```
edgibbs: Needs a newline

PR #175, opened 2018-09-19
CANS-278: Fix bug with cancel redirect

edgibbs:
Stan found an issue where canceling on an existing record with a different county selected caused
it to redirect back to the main child listing instead of the child detail.  This should fix.


app/javascript/Application/components/Client/ClientAddEditForm.js
```diff
@@ -336,10 +336,19 @@ class ClientAddEditForm extends Component {
...
 
-  redirectPath = (isNewForm, childId) => {
+  redirectPath = childId => {
     const { redirection, isUsersCounty } = this.state;
     const { successClientId } = redirection;
-    return !isUsersCounty || (isNewForm && !successClientId) ? `/` : `/clients/${childId}`;
+    const isEdit = !!childId;
+    const isCancel = !successClientId;
+    const notUsersCounty = !isUsersCounty;
+    if (isUsersCounty && isEdit) {
```
denys-davydov: Correct me if I'm wrong: when you are in edit client mode and change a county for the client, and **than** press Cancel button, your redirect path is calculated based on **new** county value, which is not the actual county for the client. If the statement is correct, you have to store the initial client's county value and base your redirect path logic for Cancel button on it.

denys-davydov: So the offline discussion shows it works as desired, resolving

PR #184, opened 2018-09-21
CANS-350 Global alert message does not stick when scrolling down 

vladimirPe:
## Description
Fix for IE11

react-stickyfill component library was added
Sticky header is wrapped with `<Sticker>` now.

## Tests

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


package.json
```diff
@@ -29,10 +29,12 @@
     "react-dom": "^16.4.1",
     "react-input-mask": "^2.0.2",
     "react-router-dom": "^4.2.2",
+    "react-stickyfill": "^0.2.3",
     "react-widgets": "^4.4.4",
     "react-widgets-moment": "^4.0.20",
     "react-wood-duck": "^0.1.57",
-    "reactstrap": "^6.0.1"
+    "reactstrap": "^6.0.1",
+    "stickyfilljs": "^2.1.0"
```
denys-davydov: Do we still need this dependency, even when we have 'react-stickyfill' already?

edgibbs: Tested it works fine and react-stickyfill does pull in the dependency anyway on stickyfill and adds it to yarn.lock

PR #194, opened 2018-09-27
CANS-400 print through keyboard

nasar-cambria:
## Description
https://osi-cwds.atlassian.net/browse/CANS-400

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -50,8 +50,32 @@ class AssessmentContainer extends Component {
     }
     this.setState({ isEditable: isEditable });
     assessmentId ? this.fetchAssessment(assessmentId) : this.fetchNewAssessment();
+    this.activeCtrlP();
   }
 
+  componentWillUnmount() {
```
edgibbs: Need to replace this, as it's a deprecated lifecycle method:

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -50,8 +50,32 @@ class AssessmentContainer extends Component {
...
+    this.muteCtrlP();
+  }
+
+  activeCtrlP = () => {
+    if (!this.state.shouldPrintNow) {
+      window.addEventListener('keydown', this.handleCtrlP, false);
```
edgibbs: Should test that you event listeners get added and removed.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -235,6 +259,7 @@ class AssessmentContainer extends Component {
     const printButton = this.renderPrintButton();
     return (
       <Fragment>
+        <div id="testAnchor" role="presentation" onKeyDown={this.handleCtrlP} />
```
edgibbs: It might be ok to have a marker for testing, but this has an onkeydown which I'm pretty sure we don't want.

app/javascript/Application/components/Assessment/AssessmentContainer.test.js
```diff
@@ -103,6 +103,28 @@ describe('<AssessmentContainer />', () => {
       });
     });
 
+    describe('print through keyboard', () => {
+      it('when a user press ctrl+p', () => {
```
edgibbs: These test are artificial, and probably don't test the behavior well enough.

app/javascript/Application/components/Assessment/AssessmentContainer.js
nasar-cambria: Please correct me if I am wrong. I see ```componentWillMount``` is deprecated and I am using  ```componentWillUnmount``` to remove the component. I see the ```unsafe_``` is also deprecated.

app/javascript/Application/components/Assessment/AssessmentContainer.js
nasar-cambria: Adding tests for Add and Remove eventListener.

app/javascript/Application/components/Assessment/AssessmentContainer.js
nasar-cambria: I am using keyDown to expect the event when it gets fired if I remove this I would not be able to capture it. One way to do this would be to refactor the <div> above which passes onClick and onKeyPress but that would be bigger change. Within the scope of this spike story I thought this can be our ideal fix to Implement and test print with formatted version.

app/javascript/Application/components/Assessment/AssessmentContainer.test.js
nasar-cambria: Behavior testing is impossible to test I tried a lot of different ways and it didn't work. So, I am testing the functionality and it test's our method successfully.

app/javascript/Application/components/Assessment/AssessmentContainer.js
edgibbs: Just removed the div and tested in 2 browsers works fine still.

app/javascript/Application/components/Assessment/AssessmentContainer.test.js
edgibbs: My point here is that adding a div with an onKeyDown to real production code is not the best approach.  I'm ok with testing the side effects at this slightly higher level as to whether the print dialog appears or not, just not with adding a div and an onKeyDown property just to make the test easier.

app/javascript/Application/components/Assessment/AssessmentContainer.js
nasar-cambria: I mean I am using it for test case not functionality. Removing div will not have any impact to the functionality. In the test case I wasn't able to find the method directly or map keydown event. So, I am using this as a marker for testing it so that I can capture the event when it gets fired.

app/javascript/Application/components/Assessment/AssessmentContainer.test.js
edgibbs: Alright let's try a short pairing session.

edgibbs: Looks good, JS testing is always fun.

PR #208, opened 2018-10-10
CANS-529 - Resolve Jest Warnings

wmitchell-cambria:
## Description

[CANS-529 - Remove warnings in test run results](https://osi-cwds.atlassian.net/browse/CANS-529)

This removes a variety of warnings and error messages from the Jest run results. While these were not causing test failures, many of them were symptomatic of real issues. For example, improper properties were passed to components, incomplete data, or wrong code paths being followed.

## Tests

## Types of changes

## Checklist:


wmitchell-cambria:
Based on the Jenkins output, it looks like there is one more error that is being output. This one only shows up temporarily in the local run. I'd still like to keep this work so far moving forward (reviewed and merged), and I will create a second PR for the additional case.

wmitchell-cambria:
retest this please

app/javascript/Application/components/Print/Print.js
```diff
@@ -32,7 +32,7 @@ class Print extends React.Component {
   print() {
     if (isFirefox) {
       this.printInFirefox();
-    } else {
+    } else if (!this.props.isTest) {
       this.printInOtherBrowser();
```
DmitriyRudenko: I wouldn't recommend to use any "test" related functionality in the actual implementation

app/javascript/Application/components/Print/Print.js
```diff
@@ -44,8 +44,13 @@ class Print extends React.Component {
 }
 
 Print.propTypes = {
+  isTest: PropTypes.bool,
```
denys-davydov: Is there a way to get rid of the error message in tests without changing the component interface? Not a fan of having test flags in code.

wmitchell-cambria:
Thanks! Additional tech debt story created here: https://osi-cwds.atlassian.net/browse/CANS-546

app/javascript/Application/components/Print/Print.js
DmitriyRudenko: how about to just re-define those functions in the test?

wmitchell-cambria: That is another possibility, although it would force the unit test to depend more on the internal implementation of the component. If mocking the functions is necessary, I think pulling them into their own module makes that more clear.

In general, code that depends on global state, like these calls to `window.focus()` or other places that set `window.location` are best isolated to their own modules. That helps keep the React components pure (in the functional sense). That approach has worked well for Intake.

PR #209, opened 2018-10-10
CANS-411 Optimize Build processes

igor-chornobay-engagepoint:
[CANS-411 Optimize Build processes](https://osi-cwds.atlassian.net/browse/CANS-411)
## Description

Split the docker build in two parts: 
- base image (not needed to be built every time. most time consuming)
- application deployment

Some changes in docker-compose for controlled startup of services
The above allowed to change sleep time in Jenkins acceptance tests from 120 s to 30 s  removing 1.5 minute of unconditional wait time

## Tests

- [x] I have included other tests
 
## Types of changes

## Checklist:


igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

docker-compose.yml
```diff
@@ -1,36 +1,47 @@
-version: '3.2'
```
money: Do we really want to go back?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+Dockerfile
```
money: This may not be needed since there never was Dockerfile in root.  Are you concerned someone might be using a local copy of Dockerfile that could get checked in?

docker-compose.yml
igor-chornobay-engagepoint: old version allowed to use depend on with condition. new one doesn't. they want you to complicate things with custom waiting scripts

.dockerignore
igor-chornobay-engagepoint: not hurting anything. let it be.

docker-compose.yml
igor-chornobay-engagepoint: there can be more work done on compose part to reduce/remove host port mappings that will reduce port conflict issue.

igor-chornobay-engagepoint:
retest this please

docker-compose.yml
edgibbs: And then most of these changes are to downgrade to version 2, this seems like a lot of change to avoid doing it the new way.

refresh_base_images
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
edgibbs: Would expect probably a package.json script hook , or m.aybe Rake

config/webpacker.yml
```diff
@@ -35,9 +35,9 @@ development:
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
     https: false
-    host: localhost
+    host: 127.0.0.1
     port: 3035
-    public: localhost:3035
+    public: 127.0.0.1:3035
```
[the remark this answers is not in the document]
igor-chornobay-engagepoint: this change is for developers using windows.

docker-compose.yml
```diff
@@ -1,36 +1,47 @@
...
     environment:
       POSTGRES_PORT: ${POSTGRES_PORT}
+    networks:
+      - cans
 
   perry:
     image: cwds/perry
     hostname: perry
+    expose:
+      - 8080
```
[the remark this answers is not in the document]
igor-chornobay-engagepoint: for ports that should not be available from outside docker-compose it is better to use expose instead of ports mapping to host. This is also a cause of port conflict issue we see in cans CI when parallel builds run. Too many ports unnecessary mapped to host causing port conflicts. We probably should have separate docker compose files for CI and for local development.
This particular change doesn't hurt anything.

Jenkinsfile
```diff
@@ -98,7 +98,7 @@ def acceptanceTestStage() {
     withEnv(["HOST=${hostname}"]) {
       withDockerRegistry([credentialsId: DOCKER_REGISTRY_CREDENTIALS_ID]) {
         sh "docker-compose up -d --build"
-        sh "sleep 120"
+        sh "sleep 60"
```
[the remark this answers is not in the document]
igor-chornobay-engagepoint: I've build this PR many times. It's not failing. The reason partially in services starting order which uses "old feature" depends_on with condition. This is main reason I moved to the older version.

docker-compose.yml
igor-chornobay-engagepoint: doing it the new way: docker recommends to use custom waiting bash scripts for controlling startup order https://docs.docker.com/compose/startup-order/
I just didn't want to complicate things for now.

.dockerignore
```diff
@@ -0,0 +1,3 @@
...
+reports
```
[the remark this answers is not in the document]
igor-chornobay-engagepoint: not hurting anything, but removed them as asked.

refresh_base_images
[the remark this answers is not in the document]
igor-chornobay-engagepoint: this is the helper script to refresh and upload to DockerHub the base images. It is intended to be run manually.  The base images should be refreshed (optionally) when the new gem(s) added or new package dependency, so the new item will be part of base image instead of being fetched and installed every time during the builds. Also should be refreshed at least every 3-6 monthe to get latest OS updates.

config/webpacker.yml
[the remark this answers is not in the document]
wmitchell-cambria: Ed mentioned to me offline that several CANS developers currently use Windows. There may be another way to do this that we were not aware of on Intake.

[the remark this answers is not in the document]
igor-chornobay-engagepoint: localhost work for windows devs. 
In intake we originally had 0.0.0.0 that presented a problem for windows devs. 
I personally don't like localhost here, but will restore my change as requested.

Jenkinsfile
```diff
@@ -98,7 +98,7 @@ def acceptanceTestStage() {
     withEnv(["HOST=${hostname}"]) {
       withDockerRegistry([credentialsId: DOCKER_REGISTRY_CREDENTIALS_ID]) {
         sh "docker-compose up -d --build"
-        sh "sleep 120"
+        sh "sleep 30"
```
igor-chornobay-engagepoint: controlled startup allowed to reduce sleep time to 30. Also worked with 20 but didn't work with 5. So leaving 30 for now

igor-chornobay-engagepoint:
image
latest build when Jenkins is not busy

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
image

!!! fastest build time so far.

igor-chornobay-engagepoint:
@abelayt @edgibbs We don't lose anything by downgrading to version 2.1. Instead we gain **easy** way to control startup order that helped to decrease the docker-compose startup time. 
The task was to optimize build time which I did by cutting it almost in half.
If we really need to have latest version of docker-compose, than it is a task for FIT team.
Please review my comments and approve.

docker-compose.yml
wmitchell-cambria: From the link you posted:
> To handle this, design your application to attempt to re-establish a connection to the database after a failure. If the application retries the connection, it can eventually connect to the database.

Sorry if I am being a newbie here, but don't our applications already do this? If we lose connection to Postgres or DB2, do we need to restart the APIs attached to it?

edgibbs: So a few things:

* As a default downgrading is not preferable
* Adding a simple check with wait-for-it or something would guarantee the service is up in a usable state and gets rid of the need for any sleeping 
* County Admin had issues where depends_on didn't operate properly for them, Kevin may be able to give more context.

docker-compose.yml
```diff
@@ -60,14 +73,33 @@ services:
       CANS_AUTHORIZATION_ENABLED: ${CANS_AUTHORIZATION_ENABLED}
       REDIS_HOST: 'redis'
       REDIS_PORT: 6379
+    depends_on:
+      redis:
+        condition: service_started
```
edgibbs: How is service_started different than just listing a depends_on: for redis?

docker-compose.yml
abelayt: > @abelayt @edgibbs We don't lose anything by downgrading to version 2.1. Instead we gain **easy** way to control startup order that helped to decrease the docker-compose startup time.
> The task was to optimize build time which I did by cutting it almost in half.
> If we really need to have latest version of docker-compose, than it is a task for FIT team.
> Please review my comments and approve.

Yes that is my understanding after I did some research and discussing with my team. The goal was to optimize the build processes and that has been achieved, hence approving this PR.

edgibbs: So experimented with migrating back to 3.2. It looks fine and the builds are just as fast after some caching on the slave.

You an see the minimal changes here in the last commit:
https://github.com/ca-cwds/cans/pull/217/commits/50233a76b3305989ca2839d856807e6d7f6d8428

igor-chornobay-engagepoint:
re-test please

igor-chornobay-engagepoint:
retest this please

igor-chornobay-engagepoint:
@edgibbs Please re-visit this PR. I did went back to version "3.2".
- Restored docker-compose to its original state before this PR
- Created new docker-compose.ci.yml using the version "3.2" which:
--- removes host port mappings (to avid port conflicts while parallel building)
--- has new service cans-test-all when run makes sure to start all dependent services and wait till they UP
- Changed Jenkins accordingly

image

refresh_base_images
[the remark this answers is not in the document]
edgibbs: Maybe make the refresh_base_images.sh so it's obvious a shell script.

docker-compose.ci.yml
```diff
@@ -0,0 +1,110 @@
...
+    depends_on:
+      - redis
+      - cans-api
+      - perry
+    tty: true
+    stdin_open: true
+    networks:
+      - cans
+
+  cans-test:
```
edgibbs: Does this need cans-test defined for CI?  I'd rather just have it all in one dockerfile, but I'm OK with the two, but cans-test probably doesn't need to be in either one now as cans-test-all has replaced it, and the other Dockerfile is just for running locally.

Jenkinsfile
```diff
@@ -97,9 +97,9 @@ def acceptanceTestStage() {
     hostname = sh(returnStdout: true, script: '/sbin/ifconfig eth0 | grep "inet addr:" | cut -d: -f2 | cut -d " " -f1').trim()
     withEnv(["HOST=${hostname}"]) {
       withDockerRegistry([credentialsId: DOCKER_REGISTRY_CREDENTIALS_ID]) {
-        sh "docker-compose up -d --build"
-        sh "sleep 120"
-        sh "docker-compose exec -T cans-test bundle exec rspec spec/acceptance"
+        sh "docker-compose -f docker-compose.ci.yml build"
+        sh "docker-compose -f docker-compose.ci.yml run cans-test-all"
```
edgibbs: Nice killed the sleep. 😺

edgibbs:
retest this please

docker-compose.ci.yml
edgibbs: After a little more staring at it, looks like it's still used.

refresh_base_images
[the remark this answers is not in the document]
igor-chornobay-engagepoint: renamed it with .sh

igor-chornobay-engagepoint:
retest this please

PR #215, opened 2018-10-12
Cans 334 implement semver versioning

zliuwd:
## Tests

## Types of changes

## Checklist:

## Remarks:
1. Most of the code comes from Devops and currently be used for some other projects
2. The only way to get this tested  is merge it firstly(temporarily), if there is any building failed, we could recover it immediately because we also keep the original Jenkinsfile in this Pr we could switch back as needed.


zliuwd:
Hi will, sorry for confusing. I answer your last question first and then I will provide some detail to the code.
The reason I did not check it as stable and won't break the build. Because the new autotag-jenkinsfile is still in testing, and it is a separate file from the original jenkinsfile, so what we gonna do here is some setting change to make autotag-jenkinsfile be used for building temporarily (may be just one time build ), during this process if any break or fail happen we will change back to original jenkinsfile immediately. By this we will see if this code could actually work for master branch.

It means during this process it could fail the build, but as we have the origin jenkinsfile in there, once it happens we will recovery it asap.

zliuwd:
Hi Will, I try to answer you first part of the question as below. Since I am still learning Groovy and Jenkins so maybe have some mistakes in my explanation. Thank you for any comment and understanding.

At first, the version number changing is based on Pr labels: Major, minor or patch, and the base version number already set in github now is 0.9.26. For example, if there is a Major change the version number could be increased to 1.9.26, if just patch the next number will be increased to 0.9.27

Secondly, it will change the github version number, dockerhub version number, Prient deploy parameter and the version number in package.json.

Thirdly, the result we expect is we could have version number updating every time when we have new Pr merged or just manually rebuild without any Pr. (There should be a version number increment after every each build )    

The Fourth, the approaching for every stages as below: 

Because Cans have three build processings separately for Pr request build, Master Build and Acceptance Build, I just use Master build stages as an example.

Master build stages as below:
checkoutStage()
buildDockerImageStage()
lintAndUnitTestStages()
acceptanceTestStage()
a11yLintStage()
publishImageStage()
deployToPreintStage()

checkoutStage contain two stages as below:
      1. Checkout stage will Hook up the github repo to make the pr event could be accessed
      2. Calculate stage :
Firstly, During this stage, we have a checker “versionIncreasment”  to check if the version number should be increased. The code will check the the Pr event to find if any “Target Labels” (Major,Minor, and Patch) in there. If the code find one Target Label in the event it will get according version number increased (based on the last version number get from github). If there are more than one Target Labels the code will return Exception to request labels change and break the build. If there are no Target Labels found then the code will return null to the checker “versionIncreasment” .
 
Secondly, The Calculate stage will  use the checker “versionIncreasment” to generate the new version number, if  versionIncreasment is null it will use the default “patch” label to generate the new version number. It means that if a Pr merged without a label or we just manually click the rebuild we will also get the version increased.

Thirdly, Also in this stage, After we got the new version number, we will change the version number in package.json

      B.buildDockerImageStage
      In this stage the code will use the new version number to build the Docker image.

      C.D.E these three stages will do some tests.

F. publishImageStage
This stage also contain three stages.
1, publish Docker image. In this stage the code will publish the Docker image with new version number to Docker hub.
2, update the git repo with new version number.
3, security scan to this Docker container with new version number

G: deployToPreintStage
In this stage the code will deploy the image to Preint with the new version number as the build parameters.
` sh "curl -u $jenkinsauth 'http://jenkins.mgmt.cwds.io:8080/job/preint/job/deploy-cans/buildWithParameters?token=deployPreint&version=${newTag}'" `

For the issue that deployed old image to the Preint, it is just because the stage G (yesterday), at that time we still use Build.ID as the deploy version number. And we already change it to newTag (as the code upwards), so I believe it will deploy the right version to Preint.

Thank you again for your comment and the help from Leonid, Alex(devops) an Ed.

wmitchell-cambria:
What happens if somebody manually pushes a tag that is not SemVer? Based on my reading of the code, it looks like the tag sorting would fail.

zliuwd:
Hi Will, You mean if a person set a Pr use the  label like "bug", "issue"...rather than the label among Major, Minor, and Patch, right?
1 . If like so, the "versionIncrement" checker will be returned as null (the logic down below), this null value will be passed to variety "increment" in stage('Calculate version')  and there is a logic if "increment" is null the calculate version stage will force this build use "patch" label then increase the last number of the version.

```
def getVersionIncrement(labels) {
    debug("getVersionIncrement( labels: ${labels} )")

    def versionIncrement = null
    def versionIncrementsFound = 0
    for(label in labels){
        switch(label) {
            case "major":
                versionIncrement = VersionIncrement.MAJOR
                versionIncrementsFound++
                break
            case "minor":
                versionIncrement = VersionIncrement.MINOR
                versionIncrementsFound++
                break
            case "patch":
                versionIncrement = VersionIncrement.PATCH
                versionIncrementsFound++
                break
        }
    }

    if(versionIncrementsFound > 1)
        throw new Exception("More than one version increment label found. Please label PR with only one of 'major', 'minor', or 'patch'")

    return versionIncrement
}
```

2. If what you means that a person manually change the version number in our git repo, I think it would impact the calculation, but there still a rescue logic for that as below (avoid building fail):
```
// Gets all the tags that match SemVer format
// Returns a list of strings (version number tags)
def getTags() {
    def gitTagOutput = sh(script: "git tag", returnStdout: true)
    def tags = gitTagOutput.split("\n").findAll{ it =~ /^\d+\.\d+\.\d+$/ }
    return tags
}
``` 
It will pick out all the tags match Semver format then do the calculation base on the last version number of these. It means if only most recently version number is manually changed it will do calculation based on the one before it.

3. If our last Docker image version number be changed manually, I think it will not impact our build, because the code will tag the Docker image with whole new tag every time.

4. IMO, this code just depends on two things
First, the last git version number (The most recently one which meets the semVer), like mentioned upwards, we have backup logic to deal with the non-semantic git version number

Second, the Pr labels. The code have logic to deal with wrong label (non-semantic-labels), no label and repeat labels.

#### What I found could be a problem is in the process the code update the package.json version number.

If there is a manually changing on package.json version number to some number different from '"version": "0.1.0"'. The code should be fail to change the version number of package.json but I am not sure if it will lead to a building break, I think probably not , the logic as below:
```
def updateFiles(newTag) {
    debug("updateFiles( newTag: ${newTag} )")
	def source = readFile file: 'package.json'
	source = source.replace('"version": "0.1.0"', '"version": \"'+newTag+'\"')
	writeFile file:'package.json', text: "$source"
}
```
Thank you for the comment!

wmitchell-cambria:
Ah, great. Yeah, I was thinking if somebody manually did a `git tag` and `git push`, which it looks like you have covered. Awesome! I'll take a closer look at the Jenkinsfile, too. Thanks for adding such detailed responses to my questions!

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
+// Used to avoid known_hosts addition, which would require each machine to have GitHub added in advance (maybe should do?)
```
edgibbs: Should probably resolve comments.

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+}
+
+def checkoutStage() {
+  stage('Checkout') {
+    cleanWs()
+    git branch: '$pull_request_event_base_ref', credentialsId: '433ac100-b3c2-4519-b4d6-207c029a103b', url: 'git@github.com:ca-cwds/cans.git'
+  }
+  stage('Calculate version'){
```
edgibbs: Why is this a nested stage inside this function?

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    try {
+      def prEvent = getPullRequestEvent()
```
edgibbs: Master builds won't have pull request events, not sure how to handle this case since it's the one that counts in terms of tagging versions and dockerhub containers.

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+  stage('Acceptance Test Preint') {
+    sh "docker-compose up -d --build cans-test"
+    sh "docker-compose exec -T --env CANS_WEB_BASE_URL=https://cans.preint.cwds.io/cans cans-test bundle exec rspec spec/acceptance"
+  }
+}
+
+def publishImageStage() {
+  stage('Publish to Dockerhub') {
+    withDockerRegistry([credentialsId: DOCKER_REGISTRY_CREDENTIALS_ID]) {
+      app.push()
```
edgibbs: shouldn't this pass the version?

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+
+    // Send notifications
+
+    slackSend channel: "#cals-api", baseUrl: 'https://hooks.slack.com/services/', tokenCredentialId: 'slackmessagetpt2', color: colorCode, message: summary
+    emailext(
+            subject: subject,
+            body: details,
+            attachLog: true,
+            recipientProviders: [[$class: 'DevelopersRecipientProvider']],
+            to: "Leonid.Marushevskiy@osi.ca.gov, Alex.Kuznetsov@osi.ca.gov"
```
edgibbs: Are these the right people to email?

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    )
+}
+// Takes an array of strings (labels)
+// Returns a VersionIncrement object
+def getVersionIncrement(labels) {
+    debug("getVersionIncrement( labels: ${labels} )")
```
edgibbs: debug statements should probably removed.

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+            subject: subject,
+            body: details,
+            attachLog: true,
+            recipientProviders: [[$class: 'DevelopersRecipientProvider']],
+            to: "Leonid.Marushevskiy@osi.ca.gov, Alex.Kuznetsov@osi.ca.gov"
+    )
+}
+// Takes an array of strings (labels)
+// Returns a VersionIncrement object
+def getVersionIncrement(labels) {
```
edgibbs: On the master build there won't be labels

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    def newTag = "${newTagMajor}.${newTagMinor}.${newTagPatch}"
+    return newTag
+}
+
+def updateFiles(newTag) {
+    debug("updateFiles( newTag: ${newTag} )")
+	def source = readFile file: 'package.json'
+	source = source.replace('"version": "0.1.0"', '"version": \"'+newTag+'\"')
```
edgibbs: won't this only work once?

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    def tags = gitTagOutput.split("\n").findAll{ it =~ /^\d+\.\d+\.\d+$/ }
+    return tags
+}
+
+// Gets a string indicating what the new tag should be in SemVer format
+// Takes a list of strings in sem
+// Returns a string with the new version tag
+def getNewTag(List tags, VersionIncrement increment) {
+    // Sort because comparator is not worked well
+    for (i=0;i<tags.size();i++)  {
```
edgibbs: Is there a cleaner way to do this, lots of complexity.

autotag-Jenkinsfile
zliuwd: Hi Ed, I will try to resolve them today, Thank you for your comments!

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    } else {
+        color = 'RED'
+        colorCode = '#FF0000'
+        details += "<p>Error message ${e.message}, stacktrace: ${e}</p>"
+        summary += "\nError message ${e.message}, stacktrace: ${e}"
+    }
+
+    // Send notifications
+
+    slackSend channel: "#cals-api", baseUrl: 'https://hooks.slack.com/services/', tokenCredentialId: 'slackmessagetpt2', color: colorCode, message: summary
```
wmitchell-cambria: There is probably a better channel than #cals-api. Maybe #cans-dev?

autotag-Jenkinsfile
```diff
@@ -0,0 +1,369 @@
...
+    emailext(
+            subject: subject,
+            body: details,
+            attachLog: true,
+            recipientProviders: [[$class: 'DevelopersRecipientProvider']],
+            to: "ed.gibbs@osi.ca.gov, Leonid.Marushevskiy@osi.ca.gov, will.mitchell@osi.ca.gov, Alex.Kuznetsov@osi.ca.gov"
```
wmitchell-cambria: Just a personal request, but I don't actually need my email listed here. I will see the Slack notification first almost every time.

autotag-Jenkinsfile
zliuwd: No, because the image already be tagged with new version number in build Docker Image stage. 
```
def buildDockerImageStage() {
  stage('Build Docker Image') {
    app = docker.build("cwds/cans:${newTag}", "-f docker/web/Dockerfile .")
  }
}
```

wmitchell-cambria:
I agree with Ed that we should discuss this.

I would be amenable to merging this into master so that we can test some of the things that Ed is asking about. (They seem to work on other projects, I guess?). My concerns are not specific to this PR, but to the overall approach and how it is being shared across projects. I expressed some of those concerns in the story itself. The only other one worth mentioning goes along with one of Ed's comments:  there is a lot of complexity here, and if it is shared across projects, we should find a way not to duplicate the whole file. There _will_ be bugs in this, and we need to be able to fix them everywhere at once or it will quickly become a mess.

So, if we need to test this in master, I'd be fine merging it...but I think a few of us are uneasy about this as a final approach.

autotag-Jenkinsfile
```diff
@@ -0,0 +1,368 @@
...
+  }
+}
+
+def acceptanceTestStage() {
+  stage('Acceptance Test') {
+    hostname = sh(returnStdout: true, script: '/sbin/ifconfig eth0 | grep "inet addr:" | cut -d: -f2 | cut -d " " -f1').trim()
+    withEnv(["HOST=${hostname}"]) {
+      withDockerRegistry([credentialsId: DOCKER_REGISTRY_CREDENTIALS_ID]) {
+        sh "docker-compose up -d --build"
+        sh "sleep 120"
```
igor-chornobay-engagepoint: there were some changes made to the original Jenkinsfile related to docker compose parts. please incorporate them here.

autotag-Jenkinsfile
```diff
@@ -0,0 +1,368 @@
...
+        sh "docker-compose exec -T cans-test bundle exec rspec spec/a11y"
+      }
+    }
+  }
+}
+
+def acceptanceTestPreintStage() {
+  stage('Acceptance Test Preint') {
+    sh "docker-compose up -d --build cans-test"
+    sh "docker-compose exec -T --env CANS_WEB_BASE_URL=https://cans.preint.cwds.io/cans cans-test bundle exec rspec spec/acceptance"
```
igor-chornobay-engagepoint: there were some changes made to the original Jenkinsfile related to docker compose parts. please incorporate them here.

autotag-Jenkinsfile
zliuwd: IMO, the reason this have a nested `stage('Checkout')` inside the checkout function is that it will create a workplace use the branch has the PR event (`$pull_request_event_base_ref` this will reference to the name the PR branch ) then do the version number calculation immediately.

autotag-Jenkinsfile
zliuwd: During the `stage('Calculate version')` it will try to read the PR event by invoke the function getPullRequestEvent firstly, but if there is no PR event it will return prEvent as its default value null. Because in the getPullRequestEvent(), prEvent has been assigned a default value null as below:

```
def getPullRequestEvent() {
    def prEvents = null
    prEvents = readJSON text: env.pull_request_event
    return prEvents
}
```
So it means if there is no Pr event or just manually rebuild it will return prEvents as null, so the code will use "patch" to build.

autotag-Jenkinsfile
zliuwd: The answer same as the answer of the comment:
 "Master builds won't have pull request events, not sure how to handle this case since it's the one that counts in terms of tagging versions and dockerhub containers."  upwards.

PR #250, opened 2018-10-31
CANS-557 Social Worker Dashboard Frontend Refactoring

zliuwd:
## Description
Front-end of Build social worker dashbaord: https://osi-cwds.atlassian.net/browse/CANS-557

## Tests

## Types of changes

## Checklist:

## Updating
The Api is deployed and it is already be used to Front-end components, Please review, thank you guys!


app/javascript/Application/components/Client/ClientCardTemplateNameCell.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 
 const ClientSocialWorkerTemplateCell = props => {
   return (
-    <a key={props.original.external_id} href={`/clients/${props.original.external_id}`}>
+    <a key={props.original.external_id} href={`/cans/clients/${props.original.external_id}`}>
```
LeonidMarushevskyi: You need to use Link from 'react-router-dom' instead of href then you don't need to hard code cans prefix similar to the example in [ClientAssessmentHistory.js](https://github.com/ca-cwds/cans/blob/master/app/javascript/Application/components/Client/ClientAssessmentHistory.js):

`      <Link to={`/clients/${props.original.external_id}`}>
        {props.value}
      </Link>`

app/javascript/Application/components/Client/Client.helper.js
```diff
@@ -16,3 +17,7 @@ export function formatClientStatus(status) {
 }
 
 export const failedFetching = { message: 'Fail to fetch data from server side!' }
+
```
LeonidMarushevskyi: Please use date functions from ../../util/dateHelper.js or if there is no existing function you can use then move your function to this dateHelper. Keep all date related utils in one place.

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
```diff
@@ -19,7 +18,7 @@ export function SocialWorkerCardTemplate(client) {
       headerStyle: {
         textAlign: 'center',
       },
-      accessor: client => isoToLocalDate(client.dob),
+      accessor: client => ClientDateHelper(client.dob),
```
LeonidMarushevskyi: Maybe just need to improve existing function to support isoToLocalDate to support undefined and null values?

app/javascript/Application/components/Client/ClientCardTemplateNameCell.js
zliuwd: done

app/javascript/Application/components/Client/Client.helper.js
zliuwd: done

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
zliuwd: done

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
```diff
@@ -1,9 +1,12 @@
-import { isoToLocalDate } from '../../util/dateHelper'
 import { formatClientName, formatClientStatus } from './Client.helper'
+import { isoToLocalDate } from '../../util/dateHelper'
 import ClientCardTemplateNameCell from './ClientCardTemplateNameCell'
 import './style.sass'
 
 export function SocialWorkerCardTemplate(client) {
+  function renderDate(datetime) {
```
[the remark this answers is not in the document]
zliuwd: The reason I don't want to touch the utils rather than wrapped it up here for our particular using for this story is: I don't think leave it "blank" should be a final approaching for the empty Date or undefined key. "Blank" should be a temporary resolution for this story, because the best way is tell the user what happen here. Additionally, "invalidate date" is from "moment" dependency plugin as a default error return, Maybe there are some parts of our app using it as a condition of logic. I don't want to impact other functionality of our app.

PR #252, opened 2018-10-31
Client's assessments fix

vladimirPe:
## Description
Small fix for undefined county

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Client/Client.test.js
```diff
@@ -149,7 +149,7 @@ describe('<Client />', () => {
           clientWrapper(emptyCountiesParams)
             .find('#client-data-counties')
             .html()
-        ).toEqual(expect.stringContaining('N/A'))
+        ).toEqual(expect.stringContaining(''))
```
wmitchell-cambria: `stringContaining('')` is probably always true. Should this just be `toEqual('')`?

vladimirPe: good catch :)

vladimirPe: but it is not equals `''` thre is a div like `"<div class=\"MuiGrid-item-2 MuiGrid-grid-xs-3-32\" id=\"client-data-counties\"><div class=\"label-text\">Counties</div></div>"`
Is it OK if I just check if element is there?

wmitchell-cambria: Hmm... In that case, you may be able to do `.text()` instead of `.html()` and verify that it equals 'Counties' exactly. Try that, and I'll check out the branch and try a couple things, too.

PR #255, opened 2018-11-02
Update Loofah to patched vulnerability

wmitchell-cambria:
## Description

No Jira story. Updates the version of loofah to a patched version, addressing a potential XSS vulnerability. See https://github.com/flavorjones/loofah/issues/154

## Tests

No tests, just a dependency update.

## Types of changes

## Checklist:


wmitchell-cambria:
Lemme check on that. I think this is OK, based on an error message I saw in Intake. I think I was misinterpreting the `Gemfile.lock` format.

wmitchell-cambria:
OK, so I think this is fine. The line I changed here is the actual bundled version of loofah that will be installed. The other lines are the dependencies declared by other modules, which break if we change them manually (that's one issue that Intake ran into). When I ran `bundle update loofah`, which is what I should have done in the first place, the only thing it did was also update nokogiri from 1.8.3 to 1.8.5 (and updated the `BUNDLED WITH` line). All other lines are the same. So... I think this is fine, but we should consider following up with a nokogiri update as well.

NagaPradeepGajula-osi:
Thanks for the clarification. 👍

PR #264, opened 2018-11-07
CANS-637: Change "Confidential" to "Discretion Needed" on non-SUD items

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Item.js
```diff
@@ -187,7 +187,7 @@ class Item extends Component {
                 naValue={this.handleNaValueSetting(rating)}
               />
             ) : null}
-            <Typography variant="title" style={{ marginRight: 15 }}>
+            <Typography variant="title" style={{ marginRight: confidential_by_default ? '4rem' : '1.5rem' }}>
```
wmitchell-cambria: Different browsers will render fonts slightly differently, so I would not assume that this will always line up as intended. If this label is supposed to take up a fixed amount of space, a better approach would be to set that width explicitly.

app/javascript/Application/components/Assessment/Item.js
```diff
@@ -99,7 +99,7 @@ class Item extends Component {
           <FormControl>
             <FormControlLabel
               onChange={this.handleConfidentialityChange}
-              label={'Confidential'}
+              label={confidential_by_default ? 'Confidential' : 'Discretion Needed'}
```
wmitchell-cambria: Is `confidential_by_default` going to always be an accurate proxy for indicating Substance Use Disorder items? I worry that this won't actually accomplish the story, or that it will be easy to introduce a regression going forward. Reading this code alone, it is not at all clear why something that is not confidential by default can't be confidential _at all._

PR #265, opened 2018-11-07
CANS-707 - Remove callback URL when logging out

wmitchell-cambria:
## Description

This callback URL created unexpected behavior when logging out and back in as a different user. For example, when a Supervisor logs out, the next user should not be directed back to the Supervisor dashboard.

## Tests

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
I am marking this a breaking change, since it does remove existing functionality of returning from where you left off when your session expires.

## Checklist:


wmitchell-cambria:
Re: your comment @denys-davydov, that the user will actually end up going back to the Dashboard page, as though they logged in from nothing. Once they click into CANS, they will be routed to their appropriate Supervisor/Social Worker dashboard.

This has a side effect of redirecting you to Preint if you are using developing locally using Preint Perry. I confirmed that Intake faces the same quirk, and it should not affect any deployed environments.

PR #266, opened 2018-11-08
CANS-636 - Default age group to false in case it is null

wmitchell-cambria:
## Description

[[CANS-636] Assessment Domains and Questions Don't Display until Age Range is Selected](https://osi-cwds.atlassian.net/browse/CANS-636) (Part 1)

This explicitly sets the default value of `under_six` to false when creating a new assessment from the instrument prototype. This is *not* the full story implementation. It is a small patch to decouple the API work from the front-end work. With this, https://github.com/ca-cwds/cans-api/pull/139 can be merged before the full implementation of hiding the assessment is completed, which will in turn make testing the remaining front-end work much easier.

In addition to the slight unit test modifications, I tested this manually against both CANS API contracts (master and my PR).

## Tests

## Types of changes

## Checklist:


eckdanny-osi:
`instrument` has a property descriptor named `prototype`? 😬

wmitchell-cambria:
Yes, but it doesn't cause any issues since it is not a constructor function. We tend to clone it rather than actually use it for prototypal inheritence, though. /shrug

denys-davydov:
Welcome to the world of Design Patterns: https://www.journaldev.com/1440/prototype-design-pattern-in-java

PR #267, opened 2018-11-08
CANS-316 change rating dropdown to radio buttons-Fixing IE11 Radio Buttons display error

zliuwd:
## Tests

## Types of changes

## Checklist:

## Remark
This Patch is used for fixing the problem that: IE11 could not display the Radio buttons and their labels at correct location. 
This problem is caused by Material UI label placement attribute, in this story we need change the label placement attribute to "start" for according the design. But the IE11 does not support this Material UI setting very well.


app/javascript/Application/components/Assessment/I18nHelper.js
```diff
@@ -20,3 +20,38 @@ export function getI18nValuesByPrefix(i18n, prefix) {
     })
   return results
 }
+
+// export const isIE = navigator.userAgent.indexOf('MSIE') != -1 || Boolean(document.documentMode) === true
```
denys-davydov: This line is probably here by mistake

app/javascript/Application/components/Assessment/I18nHelper.js
```diff
@@ -20,3 +20,38 @@ export function getI18nValuesByPrefix(i18n, prefix) {
...
+
+export function radioIeFixer(isIE) {
```
denys-davydov: It doesn't seem to be the best place for the function. I would prefer to create a separate file named for example `ItemStyling.js`.

denys-davydov: And the function name will change to `radioButtonStyle()`

app/javascript/Application/components/Assessment/I18nHelper.js
```diff
@@ -20,3 +20,38 @@ export function getI18nValuesByPrefix(i18n, prefix) {
...
+  let styleSetting
+  if (isIE === undefined) isIE = navigator.userAgent.indexOf('MSIE') !== -1 || Boolean(document.documentMode) === true
+  if (isIE) {
+    styleSetting = {
+      width: '20px',
```
denys-davydov: We are using **rem** instead of **px**, so it is a nice opportunity to fix it where you touched the wrong code. It will be `width: '2rem'`

app/javascript/Application/components/Assessment/I18nHelper.js
denys-davydov: Can't find any usage for this `isIE` parameter. You can just remove it and it will still work, so let's remove.

app/javascript/Application/components/Assessment/I18nHelper.js
```diff
@@ -20,3 +20,38 @@ export function getI18nValuesByPrefix(i18n, prefix) {
     })
   return results
 }
+
+// export const isIE = navigator.userAgent.indexOf('MSIE') != -1 || Boolean(document.documentMode) === true
+
+export function radioIeFixer(isIE) {
+  let styleSetting
+  if (isIE === undefined) isIE = navigator.userAgent.indexOf('MSIE') !== -1 || Boolean(document.documentMode) === true
```
denys-davydov: You have the same code in `labelIeFixer()` function, so factor it out as a function and reuse.

app/javascript/Application/components/Assessment/I18nHelper.js
denys-davydov: In a later comment you will find a request to factor the same code out, so you better uncomment this part and reuse it.

app/javascript/Application/components/Assessment/I18nHelper.js
```diff
@@ -20,3 +20,38 @@ export function getI18nValuesByPrefix(i18n, prefix) {
...
+    styleSetting = {
+      width: '10px',
+      '&$checked': {
+        color: '#09798E',
+      },
+    }
+  }
+  return styleSetting
+}
+export function labelIeFixer(isIE) {
```
denys-davydov: All the comments you have for the `radioIeFixer()` function are applicable to this function as well.

zliuwd:
Hi Denys, fixed all of them just keep the isIE, this is a maker for testing, thanks again for comments!

PR #270, opened 2018-11-10
CANS-603: confidential items has redacted rating

LeonidMarushevskyi:
## Description

https://osi-cwds.atlassian.net/browse/CANS-603

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Print/PrintAssessment.js
```diff
@@ -57,24 +72,13 @@ class PrintAssessment extends PureComponent {
...
-          </div>
-          <div style={flex}>
-            {item.has_na_option && this.renderOption(item.rating === 8, 'N/A')}
-            {this.renderOption(item.rating === 0, isRegularType ? '0' : 'No')}
-            {this.renderOption(item.rating === 1, isRegularType ? '1' : 'Yes')}
-            {isRegularType && this.renderOption(item.rating === 2, '2')}
-            {isRegularType && this.renderOption(item.rating === 3, '3')}
-          </div>
-        </div>
+        {!item.confidential_by_default ? this.renderOptions(item, isRegularType) : <div style={flex}/>}
```
denys-davydov: You should use `item.confidential` field instead of `item.confidential_by_default` in this logic. `confidential_by_default` is an Item property that is not changed by a user, it is a part of an instrument rather than an assessment form. At the same time `confidential` is a field that is changed by a user and the logic of hiding/showing the item rating is supposed to be based on it.

LeonidMarushevskyi: Good catch. We actually should use both flags because we want to apply logic only to SUD items. Fixed.

app/javascript/Application/components/Print/PrintAssessment.js
```diff
@@ -57,24 +73,17 @@ class PrintAssessment extends PureComponent {
...
-          </div>
-          <div style={flex}>
-            {item.has_na_option && this.renderOption(item.rating === 8, 'N/A')}
-            {this.renderOption(item.rating === 0, isRegularType ? '0' : 'No')}
-            {this.renderOption(item.rating === 1, isRegularType ? '1' : 'Yes')}
-            {isRegularType && this.renderOption(item.rating === 2, '2')}
-            {isRegularType && this.renderOption(item.rating === 3, '3')}
-          </div>
-        </div>
+        {!(item.confidential_by_default && item.confidential) ? (
```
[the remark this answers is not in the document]
LeonidMarushevskyi: @wmitchell-cambria Unit test added. We need to discuss possible refactoring of the page in order to make it more testable.

LeonidMarushevskyi:
@wmitchell-cambria @denys-davydov Comments are addressed please revisit

app/javascript/Application/components/Print/PrintAssessment.test.js
```diff
@@ -4,13 +4,29 @@ import PrintAssessment from './PrintAssessment'
...
+    const actualHtml = printAssessmentHtml.replace(
       /<span style="text-align:right;font-style:italic">[^>]+<\/span>/g,
       '<span style="text-align:right;font-style:italic" />'
     )
     expect(actualHtml).toEqual(printFixture)
   })
+
+  it('should render confidential item and redact rating when confidential by default and confidential is true', () => {
+    expect(printAssessmentHtml).not.toContain('Confidential')
+    expect(printAssessmentHtml).not.toContain('<div style="width:7.6rem"></div>') // Redacted rating
```
wmitchell-cambria: This seems very fragile. If somebody changes the style, for example, this will silently pass even if it is not redacting the ratings.

app/javascript/Application/components/Print/PrintAssessment.test.js
```diff
@@ -4,13 +4,29 @@ import PrintAssessment from './PrintAssessment'
...
+    assessmentPrint.state.domains[0].items[0].confidential_by_default = true
```
wmitchell-cambria: This modifies the assessmentPrint object for any following tests. Consider cloning it or creating an assessment object with a single Item, form-fit for this test. That will also be clearer than using a large assessment object with a mix of ratings.

app/javascript/Application/components/Print/PrintAssessment.test.js
```diff
@@ -4,13 +4,29 @@ import PrintAssessment from './PrintAssessment'
...
+    assessmentPrint.state.domains[0].items[0].confidential = true
+    const printConfidential = shallow(<PrintAssessment assessment={assessmentPrint} i18n={i18nPrint} />)
```
wmitchell-cambria: Multiple renders is an indication that this could be split into smaller tests that each have a single way to fail.

wmitchell-cambria:
To clarify, by 'switching to shallow rendering', I mean removing the `.html()` call, which forces a full render of the tree.

app/javascript/Application/components/Print/PrintAssessment.test.js
LeonidMarushevskyi: Thanks for the suggestion. Added small fixture with on item to test confidential logic.

LeonidMarushevskyi:
@wmitchell-cambria refactored unit test according to the recommendation. Please revisit.

app/javascript/Application/components/Print/PrintAssessment.test.js
```diff
@@ -13,4 +18,11 @@ describe('<PrintAssessment />', () => {
     )
     expect(actualHtml).toEqual(printFixture)
   })
+
+  it('should render confidential item and redact rating when confidential by default and confidential is true', () => {
+    const printConfidential = shallow(<PrintAssessment assessment={assessmentWithConfidentialItem} i18n={i18nPrint} />)
+    const printConfidentialHtml = printConfidential.html()
+    expect(printConfidentialHtml).toContain('Confidential')
+    expect(printConfidentialHtml).toContain('<div style="width:7.6rem"></div>') // Redacted rating
```
wmitchell-cambria: This is still a very brittle line, fwiw.

PR #271, opened 2018-11-11
CANS-594-Add-total-domain-score

zliuwd:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/DomainCaregiverControls.test.js
```diff
@@ -0,0 +1,71 @@
...
+  it('second li have text # + ADD CAREGIVER #', () => {
+    expect(
+      wrapper
+        .find('li')
+        .at(1)
+        .text()
+    ).toBe('+ ADD CAREGIVER')
+  })
+
+  it('will invoke onRemoveCaregiverDomain when first li be clicked or have key pressed', () => {
```
wmitchell-cambria: "or" is an indication that these should be separate test cases.

app/javascript/Application/components/Assessment/DomainItemList.test.js
```diff
@@ -0,0 +1,55 @@
...
+  let wrapper
+  beforeEach(() => {
+    wrapper = mount(<DomainItemList {...fakeItemListProps} />)
+  })
+
+  afterEach(() => {
+    wrapper.unmount()
+  })
+
+  it('will render 2 div', () => {
```
wmitchell-cambria: Ed linked an interesting article recently about writing unit tests that are immediately _obvious_. (https://mtlynch.io/good-developers-bad-tests/)

This test is not so obvious. If this test fails, the developer will read the results and see "<DomainList/> will render 2 div". It doesn't say why it would render 2 divs (and in many cases, it should render more or less than 2 divs).

Instead, consider a name like "`it('will render a div for each Item')`. That immediately explains why it should be 2 in this scenario, and describes the overall behavior much better.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -5,15 +5,13 @@ import ExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary'
...
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
 import { Input } from 'reactstrap'
-import { DomainProgressBar, Item } from './'
+import { DomainProgressBar, DomainScore, DomainItemList, DomainCaregiverControls } from './'
 import { shouldDomainBeRendered } from './AssessmentHelper'
-import { getI18nByCode } from './I18nHelper'
 import { isA11yAllowedInput } from '../../util/events'
 import Grid from '@material-ui/core/Grid'
-
+import { totalScoreCauculation } from './DomainScoreHelper.js'
```
wmitchell-cambria: Typo in this helper method name. `calculation`?

app/javascript/Application/components/Assessment/DomainScore.test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import DomainScore from './DomainScore'
+import { Badge } from '@cwds/components'
+
+describe('<DomainScore/>', () => {
+  const fakeDSNumberProps = {
+    totalScore: 5,
+  }
+  const fakeDSSringProps = {
```
wmitchell-cambria: Typo `string`

app/javascript/Application/components/Assessment/DomainScoreHelper.js
```diff
@@ -0,0 +1,15 @@
+export function totalScoreCauculation(items) {
+  let noneTouched = 0
+  let result = 0
+  items.forEach(item => {
+    let rating = item.rating
+    if (rating < 0) {
+      noneTouched++
+      rating = 0
+    } else if (rating > 3) {
```
wmitchell-cambria: Magic number.

app/javascript/Application/components/Assessment/assessment.mocks.test.js
```diff
@@ -678,8 +678,7 @@ export const i18nPrint = {
...
..."font-size:1rem">T03. Emotional Abuse</div></div><div style="display:flex;flex-direction:row"><div style="flex-grow:1;flex-shrink:1;display:flex;flex-direction:column;justify-content:center;margin:0.2rem 0.2rem"><div style="font-size:0.8rem;font-weight:400;text-align:center">Confidential</div></div><div style="flex-grow:1;flex-shrink:1;display:flex;flex-direction:column;justify-content:center;margin:0.2rem 0.2rem"><input type="checkbox" readonly=""/></div></div><div style="display:flex;flex-direction:row"><div style="flex-grow:1;flex-shrink:1;display:flex;flex-direction:column;justify-content:center;margin:0.2rem 0.2rem"><div style="font-size:0.8rem;font-weight:400;text-align:center">Rating:</div></div><div style="display:flex;flex-direction:row"><div style="display:inline-flex;flex-direction:column"><input type="radio" style="margin:0.2rem 0.2rem" checked="" readonly=""/><span style="font-size:0.8rem;font-weight:400;text-align:center">No</span></div><div style="display:inline-flex;flex-direction:column"><input type="radio" style="margin:0.2rem 0.2rem" readonly=""/><span style="font-size:0.8rem;font-weight:400;text-align:center">Yes</span></div></div></div></div></div></div></div>'
```
wmitchell-cambria: Similar to my comment on Leonid's PR - Can we add tests here to help capture some more of the actual changes? This big blob is really hard to review. The other side of it here is that it will almost certainly conflict with the changes in two different PRs.

app/javascript/Application/components/Assessment/DomainItemList.js
```diff
@@ -0,0 +1,44 @@
...
+          onRatingUpdate={props.onRatingUpdate}
+          onConfidentialityUpdate={props.onConfidentialityUpdate}
+          isAssessmentUnderSix={props.isAssessmentUnderSix}
+          canReleaseConfidentialInfo={props.canReleaseConfidentialInfo}
+        />
+        <Divider />
+      </div>
+    )
+  })
+  return <Fragment>{itemlist}</Fragment>
```
wmitchell-cambria: One thing to note is that I *think* the Fragment here is unnecessary. Since each item in the array already has a unique key, DomainItemList can return an array of children directly. Although, if we do that, note that using the array index as a key is an antipattern, and we should use something unique about the item, like its code, instead. (That helps React figure out if the items get reordered.)

Anyways, something to test out, because I am not 100% sure of the differences between returning an array vs a Fragment.

zliuwd: After test, for here, we need a Fragment. Otherwise the component will not be rendered properly. And I changed the key to code.

app/javascript/Application/components/Assessment/DomainItemList.test.js
zliuwd: Learned! Thank you!

app/javascript/Application/components/Assessment/DomainScoreHelper.js
zliuwd: Yeah! I should add some const here.

PR #272, opened 2018-11-13
CANS-663. Referral/Case Refactored

denys-davydov:
## Description
https://osi-cwds.atlassian.net/browse/CANS-663

## Tests

## Types of changes

## Checklist:

### DIFF
--- a/app/javascript/Application/components/Assessment/AssessmentContainer.js
+++ b/app/javascript/Application/components/Assessment/AssessmentContainer.js
@@ -113,10 +113,13 @@ class AssessmentContainer extends Component {
   }
 
   onFetchNewAssessmentSuccess(instrument) {
+    const client = this.props.client
     const assessment = {
       instrument_id: instrument.id,
-      person: this.props.client,
-      county: this.props.client.county,
+      person: client,
+      case_or_referral_id: client.case_or_referral_id,
+      case_or_referral_uiid: client.case_or_referral_uiid,
+      county: client.county,
       assessment_type: AssessmentType.initial,
       status: AssessmentStatus.inProgress,
       state: { ...instrument.prototype, under_six: false },
--- a/app/javascript/Application/components/Client/Client.test.js
+++ b/app/javascript/Application/components/Client/Client.test.js
@@ -89,11 +89,11 @@ describe('<Client />', () => {
     const getLength = component => wrapper.find(component).length
 
     it('renders with 12 <Grid /> components', () => {
-      expect(getLength(Grid)).toBe(12)
+      expect(getLength(Grid)).toBe(11)
     })
 
     it('renders with 9 ".label-text" styled elements', () => {
-      expect(getLength('.label-text')).toBe(9)
+      expect(getLength('.label-text')).toBe(8)
     })
 
     it('does not render No Child Data Found', () => {
--- a/app/javascript/Application/components/Assessment/AssessmentContainer.test.js
+++ b/app/javascript/Application/components/Assessment/AssessmentContainer.test.js
@@ -377,7 +377,11 @@ describe('<AssessmentContainer />', () => {
         expect(wrapper.state('assessment').instrument_id).toBeFalsy()
         wrapper.instance().onFetchNewAssessmentSuccess(instrument)
         const assessment = wrapper.state('assessment')
-        expect(assessment).toEqual(initialAssessment)
+        expect(assessment).toEqual({
+          ...initialAssessment,
+          case_or_referral_id: 'C6vN5DG0Aq',
+          case_or_referral_uiid: '0687-9473-7673-8000672',
+        })
         expect(assessment.person).toEqual(childInfoJson)
         expect(assessment.county).toEqual(childInfoJson.county)
       })
--- a/app/javascript/Application/components/Assessment/AssessmentFormHeader.js
+++ b/app/javascript/Application/components/Assessment/AssessmentFormHeader.js
@@ -107,15 +107,13 @@ class AssessmentFormHeader extends PureComponent {
   }
 
   renderCaseNumber() {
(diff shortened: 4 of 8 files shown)


app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -113,10 +113,13 @@ class AssessmentContainer extends Component {
   }
 
   onFetchNewAssessmentSuccess(instrument) {
+    const client = this.props.client
     const assessment = {
       instrument_id: instrument.id,
-      person: this.props.client,
-      county: this.props.client.county,
+      person: client,
+      case_or_referral_id: client.case_or_referral_id,
```
wmitchell-cambria: I still have concerns about this data model. By storing the ID alone, without any indication of whether it is for a case or for a referral, aren't we making it incredibly difficult to separate them out later? If we ever get a request for the user to be able to tell which is which, what is our strategy for doing that?

One way that Intake has solved this problem is by storing a full 'legacy descriptor' for some fields, like addresses, which also includes the source table of the data. That way, we have a single field for the id, but can still tell whether it came from the address table or placement home table if needed. I think a similar approach would work here as well.

Alternatively, we could have separate fields for case and referral and choose to display one over the other if necessary. This is also easy to handle on the front-end, easy to query for on the back-end, and provides the same flexibility to separate referrals and cases if needed.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -113,10 +113,13 @@ class AssessmentContainer extends Component {
...
+      case_or_referral_uiid: client.case_or_referral_uiid,
```
wmitchell-cambria: Does this `case_or_referral_uuid` capture the same information as the `case_or_referral_id` but formatted differently?

app/javascript/Application/components/Assessment/AssessmentContainer.js
denys-davydov: I've had the same concern and was told by the PO, the Service Manager, and one SME, that we will not need it. Technically it is not a big deal to do it though.

abelayt: I briefly discuss this 'case_id or referral_id' with Julie and my understanding is these two numbers are never the same and once a case number is created the referral number almost never referred other than stored on the process of client's history. So I asked her would it be much better to remove the or part  to save one of these on the same variable keep these separate like case_id and refferal_id and she said she will talk to Sonaltra about this. I think for now we should move forward with what we are given about these two numbers and demand more clarity on this from SME.

app/javascript/Application/components/Assessment/AssessmentContainer.js
denys-davydov: Yes, `case_or_referral_id` is a 10-symbol base 62 legacy key, while `case_or_referral_uiid` is its 19-digit UI representation. The end user will always see the `case_or_referral_uiid` and will never see the `case_or_referral_id`. The `case_or_referral_id` is used on the back end and in db. The `case_or_referral_uiid`is converted from the `case_or_referral_id` on the back end by this algorithm:
https://github.com/ca-cwds/api-core/blob/574c107fbe3dc70c62950c09b3e1aecfceb8e770/api-core-cms/src/main/java/gov/ca/cwds/data/persistence/cms/CmsKeyIdGenerator.java#L422

app/javascript/Application/components/Assessment/AssessmentContainer.js
wmitchell-cambria: I don't think we should move forward with these. As soon as we start using it, we can no longer separate out the data stored in this variable. If it's not a big deal, I would suggest we use two variables internally. It doesn't have to affect the display at all.

abelayt: I think we can tell the number if it is  a case or referral like,
case_exists? case_id : referral_id

denys-davydov: Identifiers for cases and referrals together are unique so it is possible to distinguish between them in future so I would prefer to leave it as is in accordance with YAGNI principle. But the PO changed mind on the issue and thinks that it is possible that the end user will benefit from having a case/referral flag in foreseeable future, so we will keep the flag on the back end to make it easier to implement in future with no debt. New commit covers this functionality.

denys-davydov:
@wmitchell-cambria Please revisit the PR

app/javascript/Application/components/Client/Client.test.js
```diff
@@ -89,11 +89,11 @@ describe('<Client />', () => {
     const getLength = component => wrapper.find(component).length
 
     it('renders with 12 <Grid /> components', () => {
-      expect(getLength(Grid)).toBe(12)
+      expect(getLength(Grid)).toBe(11)
```
wmitchell-cambria: Test description says 12.

app/javascript/Application/components/Client/Client.test.js
```diff
@@ -89,11 +89,11 @@ describe('<Client />', () => {
...
     })
 
     it('renders with 9 ".label-text" styled elements', () => {
-      expect(getLength('.label-text')).toBe(9)
+      expect(getLength('.label-text')).toBe(8)
```
wmitchell-cambria: Test description says 9.

app/javascript/Application/components/Client/Client.test.js
denys-davydov: Fixed

app/javascript/Application/components/Client/Client.test.js
denys-davydov: Fixed

PR #274, opened 2018-11-14
Cans 579 implement snapshot search

ggomez-cambriasolutions:
## Description
As a CANS User, I need a way to search for and select an existing client in CWS/CMS, so that I can find who we are looking for in the CANS tool.

## Tests

## Types of changes

## Checklist:


ggomez-cambriasolutions:
jenkins retest this please

app/controllers/api/people_searches_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+                 .search(search_params,
+                         request.uuid,
+                         security_token: session[:security_token])
+      render json: response
+    end
+
+    private
+
+    def search_params
+      params.permit(:search_term, :is_client_only, :search_after)
```
wmitchell-cambria: If you haven't already, check in with Money's changes here to make sure that pagination works as intended: https://github.com/ca-cwds/dora_api/pull/4/files#diff-04c6e90faac2675aa89e2176d2eec7d8R63

ggomez-cambriasolutions: We're not including pagination this sprint. Good call, though.

money: @ggomez-cambriasolutions you should just remove `:search_after`

money:
Ruby on Rails part looks good

app/javascript/Application/components/common/search/SearchPagination.js
```diff
@@ -0,0 +1,93 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+const NUM_RESULTS_PER_PAGE = 10
+
+class SearchPagination extends Component {
```
wmitchell-cambria: If pagination is out of scope, I would also recommend removing this component. That way, we can review it in the context of working pagination once it is added. Reviewing it now is likely to miss something, and we're carrying around the unused code in the meantime.

app/javascript/Application/components/common/search/SearchPagination.js
```diff
@@ -0,0 +1,93 @@
...
+  constructor(props) {
+    super(props)
+
+    this.totalPages = this.calculateTotalPages(this.props.totalResults, NUM_RESULTS_PER_PAGE)
```
wmitchell-cambria: It seems error prone to store this directly on `this` if it is already included in the state. Updating `this.totalPages` directly will not trigger a re-render, for example.

app/javascript/Application/components/common/search/style.sass
```diff
@@ -0,0 +1,116 @@
...
+        left: 1.5rem
+        position: absolute
+        width: 11rem
+    .information-flag 
+        background-color: $transparent-black
+        color: $white
+        padding: 0 .5rem
+        text-align: center
+    .search-item-header
+        color: #a0a0a0
```
wmitchell-cambria: Variable?

app/javascript/Application/components/common/search/style.sass
```diff
@@ -0,0 +1,116 @@
...
+        width: 19rem
+        margin: 0 auto
+        .pagination-wrapper
+            display: inline-block
+            min-width: 10rem
+            button
+                padding: .75rem 1.25rem
+                border: .1rem solid $medium-gray
+                color: $light-blue
+                background-color: #f9f9f9
```
wmitchell-cambria: Hardcoded color here as well.

app/javascript/Application/components/common/search/style.sass
```diff
@@ -0,0 +1,116 @@
...
+                padding-top: .55rem
+                padding-bottom: .55rem
+                font-size: 1.8rem
+                border-radius: .4rem 0 0 .4rem
+                border-right: 0
+            button.prev-page:disabled
+                i.fa-caret-left
+                    color: $dark-gray
+            button.next-page
+                position: relative
```
wmitchell-cambria: It seems like we are redefining a lot of styles here, duplicating styles both within the file and probably with existing pagination widgets. Is there nothing else we can share?

(Another alternative solution is to leave the pagination off entirely, until it is in scope of the story.)

app/javascript/Application/enums/AddressType.js
```diff
@@ -0,0 +1,15 @@
...
+  'Home',
+  'Homeless',
+  'Other',
+  'Penal Institution',
+  'Permanent Mailing Address',
+  'Residence 2',
+  'Work',
+])
+
+export const RESIDENCE_TYPE = '32'
```
wmitchell-cambria: Note that Snapshot is currently dealing with the issue of "We want some types other than 32 to show up."

app/javascript/Application/enums/SystemCodes.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const addressTypes = [{ code: RESIDENCE_TYPE, value: 'Home' }]
+
+export const unableToDetermineCodes = [
+  { code: 'A', value: 'Abandoned' },
+  { code: 'I', value: 'Unknown' },
+  { code: 'K', value: 'Unknown' },
+]
+
+export const systemCodes = {
```
wmitchell-cambria: It seems like there are a lot of other things that look for system codes. Are we sure we are OK with just this small subset?

app/javascript/Application/util/authorization.js
```diff
@@ -0,0 +1,23 @@
+const isUserInfoReady = (userInfo, hasAddSensitivePerson) =>
+  userInfo && (hasAddSensitivePerson === true || hasAddSensitivePerson === false)
+
+const areDifferentCounties = (userCounty, clientCounty) => userCounty && clientCounty && userCounty !== clientCounty
+
+const ALLOWED_CLIENT_COUNTIES = 1
+
+export const canUserAddClient = (userInfo, hasAddSensitivePerson, client, hasOverride) => {
```
wmitchell-cambria: It doesn't look like this is used anywhere. Should it be? (If it isn't used, can we remove it?)

app/javascript/Application/util/formatters.js
```diff
@@ -20,6 +23,36 @@ export function addTrailingSlash(url) {
...
+  const suffix = highlightedSuffix.replace(/<\/?em>/gi, '')
+  const formattedSuffix = formatNameSuffix(suffix)
+  if (!formattedSuffix) {
+    return formattedSuffix
+  }
+  const rehighlightedSuffix = suffix === highlightedSuffix ? formattedSuffix : `<em>${formattedSuffix}</em>`
+
+  return rehighlightedSuffix
+}
+export const phoneNumberFormatter = phone => {
```
wmitchell-cambria: I would definitely keep these search formatters isolated to the `Search` folder, unless you want to do the work of rectifying `phoneNumberFormatter` with the nearly identical but slightly different function `formatPhoneNumber` below.

The above formatters, with all of their highlighting behavior, should be kept with Search regardless, in my opinion.

app/javascript/Application/util/isPlacementHome.js
```diff
@@ -0,0 +1,6 @@
+export const isPlacementHome = address => {
+  if (!address) {
+    return false
+  }
+  return address.getIn(['legacy_descriptor', 'legacy_table_name']) === 'PLC_HM_T'
```
wmitchell-cambria: Same here... I'd put this in Search because, while it sounds super common and useful, we don't actually use the legacy_descriptor format anywhere else in the CANS data model right now. To reduce how many different data models we have to think about at any given time, let's keep separate data models in their own domains (aka, directories).

app/javascript/Application/util/maybe.js
```diff
@@ -0,0 +1,146 @@
+/*
+  A Maybe represents an optional value. A Maybe is either Something (if it
+  contains a value) or Nothing (if it contains null or undefined). By using
+  Maybes, we can define functions that explicitly declare that they may not
+  return a value, and we can use those values without littering null checks
+  across our codebase.
+
+  Check out the test suites for individual methods below for some handy usages.
+*/
+export class Maybe {
```
wmitchell-cambria: :heart: Maybe4Life :heart:

config/routes.rb
```diff
@@ -10,8 +10,9 @@
       end
     end
 
-    resources :clients, only: [:show] do
-    end
+    resources :people_searches, only: :index
```
wmitchell-cambria: This could probably have a more RESTful name. Not critical, though.

ggomez-cambriasolutions:
jenkins please retest this

ggomez-cambriasolutions:
jenkins retest this please

app/javascript/Application/components/Search/SearchAssessmentHistory.js
```diff
@@ -33,16 +33,21 @@ class SearchAssessmentHistory extends Component {
...
 
   renderAssessments = (assessments, fetchStatus) => {
-    return fetchStatus === LoadingState.ready && assessments.length === 0 ? (
-      <div id="no-data">No assessments currently exist for the clients.</div>
-    ) : (
-      assessments.map(assessment => <SearchAssessmentHistoryRecord assessment={assessment} key={assessment.id} />)
-    )
+    if (fetchStatus === LoadingState.ready && assessments.length === 0) {
+      return <div id="no-data">No assessments currently exist for the clients.</div>
+    } else if (fetchStatus === LoadingState.error) {
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: ok, good point.

app/javascript/Application/components/Search/SearchAssessmentHistoryRecord.js
```diff
@@ -13,7 +13,7 @@ class SearchAssessmentHistoryRecord extends PureComponent {
     return (
       <Container className={'search-history-item'}>
         <Row>
-          <Col xs="1">{<AssessmentRecordIcon status={status} />}</Col>
+          <div className="assessment-record-icon">{<AssessmentRecordIcon status={status} />}</div>
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: Yes, I changed it to allow me to customize the height/width of the icon. But I reverted this change and it looks fine.

app/javascript/Application/components/Search/SearchHelper.js
```diff
@@ -0,0 +1,82 @@
+import { Map, List } from 'immutable'
+import {
+  selectAddressTypes,
+  selectUnableToDetermineCodes,
+  systemCodeDisplayValue,
+} from '../../selectors/systemCodeSelectors'
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: ok, that makes sense.

app/javascript/Application/components/Search/SearchHelper.js
```diff
@@ -0,0 +1,82 @@
...
+import { RESIDENCE_TYPE } from '../../enums/AddressType'
+import { isPlacementHome } from '../../util/isPlacementHome'
+import { zipFormatter } from '../../util/zipFormatter'
+import { Maybe } from '../../util/maybe'
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: sounds good.

app/javascript/Application/util/authorization.js
ggomez-cambriasolutions: yes, we can remove it.

app/javascript/Application/util/formatters.js
ggomez-cambriasolutions: putting these methods in the search dir.

app/javascript/Application/components/Search/Search.service.js
```diff
@@ -0,0 +1,17 @@
+import { apiGet } from '../../App.api'
+import { encodedSearchAfterParams } from './SearchHelper'
+
+export class SearchService {
+  static getClients({ searchTerm, sortAfter = null }) {
+    let saps = []
+    let endpoint = ''
+    if (sortAfter) {
+      saps = encodedSearchAfterParams(sortAfter)
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: ok, I've removed it for now. when we get pagination working in the GEM, I'll put it back in.

app/javascript/Application/components/Search/Debouncer.js
```diff
@@ -0,0 +1,59 @@
...
+      results: [],
+      totalPages: 0,
+      totalResults: 0,
+    })
+  }
+  render() {
+    const { results, totalPages, totalResults } = this.state
+
+    return React.cloneElement(this.props.children, {
+      debounce: this.debounce,
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: ok makes sense to me. We're now passing an onChange prop to the cloned Autocompleter component.

app/javascript/Application/components/Layout/Page.test.js
```diff
@@ -35,16 +25,13 @@ describe('<Page />', () => {
       expect(breadCrumbsBuilder.length).toBe(1)
     })
 
-    it('splits sidebar and main content 3:9', async () => {
+    it('renders main content 12 columns wide', async () => {
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: We have the sidebar back on all pages, except for the client search page.

app/javascript/Application/components/common/search/AddressInfo.test.js
```diff
@@ -0,0 +1,34 @@
...
+        const wrapper = shallow(<AddressInfo {...props} />)
+        expect(wrapper.find('span').exists()).toBe(true)
+      })
+    })
+  })
+
+  describe('renders address info', () => {
+    it('renders the address', () => {
+      const wrapper = shallow(<AddressInfo {...props} />)
+      expect(wrapper.find('span').text()).toEqual('1234 Hansons Way Sacramento, CA 95834')
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: we've added test case that validate address info when those pieces of info are missing.

app/javascript/Application/components/common/search/FullName.js
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types'
+import sanitizeHtml from 'sanitize-html'
+
+const FullName = ({ fullName }) => {
+  const sanitizedField = field => ({
+    dangerouslySetInnerHTML: {
+      __html: boldMatchedTerm(sanitizeHtml(field, { allowedTags: ['em'] })),
+    },
+  })
+  const boldMatchedTerm = field => field.split('em').join('strong')
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: good catch, Will! We are now bolding matched fields through css.

app/javascript/Application/components/common/search/Gender.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Genders from '../../../enums/Genders'
+
+const Gender = ({ gender }) => {
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: done.

app/javascript/Application/components/common/search/LegacyInfo.test.js
```diff
@@ -0,0 +1,33 @@
...
+        const wrapper = shallow(<LegacyInfo legacyId={props.legacy_id} />)
+        expect(wrapper.find('div').length).toBe(1)
+      })
+    })
+  })
+
+  describe('LegacyInfo', () => {
+    it('renders the legacy ui id', () => {
+      const wrapper = shallow(<LegacyInfo legacyId={props.legacy_id} />)
+      expect(wrapper.html()).toContain('<div>9QgejQ80Py</div>')
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: I wasn't sure which id to use in this case, but the more user-friendly ID (the 19 digit version) is what I'll use.

app/javascript/Application/components/common/search/SearchPagination.js
ggomez-cambriasolutions: alright, sounds good.

app/javascript/Application/components/common/search/SearchPagination.js
ggomez-cambriasolutions: yeah makes sense.

app/javascript/Application/components/common/search/style.sass
ggomez-cambriasolutions: done.

app/javascript/Application/components/common/search/style.sass
ggomez-cambriasolutions: n/a

app/javascript/Application/components/common/search/style.sass
ggomez-cambriasolutions: Leaving the search pagination off for now. When we add it back it, I'll make sure to optimize the search pagination styles.

app/javascript/Application/enums/AddressType.js
ggomez-cambriasolutions: We're not displaying the type of address in search results, so this might not be a concern for CANS.

app/javascript/Application/enums/SystemCodes.js
ggomez-cambriasolutions: those are all the system codes that I needed for client search.

ggomez-cambriasolutions:
jenkins retest

ggomez-cambriasolutions:
jenkins retest this please

ggomez-cambriasolutions:
jenkins retest this please

app/javascript/Application/components/Search/Autocomplete/PersonSuggestion.js
```diff
@@ -0,0 +1,94 @@
...
+  gender,
+  address,
+  languages,
+  isSensitive,
+  isSealed,
+  legacyDescriptor,
+  clientCounties,
+}) => {
+  return (
+    <div className="row">
```
wmitchell-cambria: One thing, while I think of it, is that we do have Row and Col components in CANS, which would be better than setting the class names directly. But...since this is copied from elsewhere, I would not recommend updating it in this PR. Possibly a future tech-debt topic.

app/javascript/Application/components/Search/Autocomplete/SearchResults/Gender.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Genders from '../../../../enums/Genders'
+
+const Gender = ({ gender }) => {
+  const genderInfo = [Genders[gender]].filter(Boolean)
+  return genderInfo[0] ? <div>{genderInfo[0]}</div> : null
```
wmitchell-cambria: Alternatively, could this just be `Genders[gender] ? <div>{Genders[gender]}</div> : null`?

Constructing an array just to filter it seems a little odd. Although, shout out to Maybe here to represent an optional value. With Maybe, it would look like this:

```javascript
Maybe.of(Genders[gender])
  .map((genderInfo) => <div>{genderInfo}</div>)
  .valueOrElse(null)
```

Now, I'm not saying that is cleaner than the ternary... but I am saying it's cooler.

app/javascript/Application/components/Search/Autocomplete/SearchResults/ClientDob.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const ClientDob = ({ dateOfBirth, sanitizedDob, dob }) => {
+  if (dateOfBirth === `<em>${sanitizedDob}</em>`) {
+    return (
+      <div>
+        <span className="client-search-matched-field">{dob.format('M/D/YYYY')}</span>
```
wmitchell-cambria: If we are already styling `<em>` to be light blue and bold, then could we just render dateOfBirth directly? Based on the query, I worry that there are cases where we match month and day but not year, for example.

app/javascript/Application/components/Search/Autocomplete/SearchResults/ClientDob.js
```diff
@@ -0,0 +1,37 @@
...
+
+ClientDob.defaultProps = {
+  dateOfBirth: '',
+  sanitizedDob: '',
+  dob: {},
+}
+ClientDob.propTypes = {
+  dateOfBirth: PropTypes.string,
+  dob: PropTypes.object,
+  sanitizedDob: PropTypes.string,
```
wmitchell-cambria: The purposes of these three props are very hard to tell apart. :-)

app/javascript/Application/components/Search/Autocomplete/SearchResults/LanguageInfo.js
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+export const MAX_LANGUAGES = 2
+
+export const flagPrimaryLanguage = languages => {
+  if (languages[0]) languages[0] += ' (Primary)'
```
wmitchell-cambria: This mutates the array passed in. Can this return a new array that doesn't modify the original? Or, if it must modify the original, let's not have it return anything, so that the consumer doesn't accidentally think they are getting a new array.

app/javascript/Application/components/Search/Autocomplete/SearchResults/LanguageInfo.js
```diff
@@ -0,0 +1,30 @@
...
+  return languages
+}
+
+export const primaryLanguageOnly = languages => {
+  return languages[0] ? languages[0] : ''
+}
+
+const LanguageInfo = ({ languages }) => {
+  const lan = languages && primaryLanguageOnly(languages.filter(Boolean))
```
wmitchell-cambria: Based on the propTypes, we always expect `languages` to be truthy here. If we need this guard, it means we cannot expect that, so we should change the propTypes.

app/javascript/Application/components/Search/SearchAssessmentHistory.test.js
```diff
@@ -104,7 +105,7 @@ describe('<SearchAssessmentHistory', () => {
         ])
 
         // then
-        expect(wrapper.find(SearchAssessmentHistoryRecord).length).toBe(3)
```
wmitchell-cambria: For future record, you can safely ignore 'magic number' warnings in test files. We need to get different lint rules set up for test files soon.

app/javascript/Application/components/Search/SearchHelper.js
```diff
@@ -0,0 +1,82 @@
...
+      .map(race => race.get('description'))
+  } else {
+    return List()
+  }
+}
+const isResidence = address => address.getIn(['type', 'id']) === RESIDENCE_TYPE
+const addressMaybe = result =>
+  Maybe.of(result.get('addresses'))
+    .map(as => as.first())
+    .filter(isResidence)
```
wmitchell-cambria: We filter based on residence type here.

app/javascript/Application/enums/AddressType.js
wmitchell-cambria: It is used to filter out the addresses that are displayed, even if we aren't displaying the type.

app/javascript/Application/enums/SystemCodes.js
wmitchell-cambria: In that case, I think it would be worthwhile to remove the `systemCodeSelector` functions and mocked values for the other code types. Having the codes here is definitely a form of tech debt, since any developer in this directory has to start worrying about language codes, for example, even when they aren't relevant.

app/models/county_staff.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CountyStaff
```
wmitchell-cambria: This file is unused, I think.

app/javascript/Application/components/Search/Autocomplete/SearchResults/Gender.js
ggomez-cambriasolutions: like totally...

app/javascript/Application/components/Search/Autocomplete/SearchResults/ClientDob.js
ggomez-cambriasolutions: this one is a little strange, but it's checking whether the entire DOB should be bolded, just a substring of DOB, or should render the DOB not bold. That is why we don't render just the `<em>${sanitizedDob}</em>`. I don't think we should change this one. But if you still think so, then maybe you can explain a little more.

app/javascript/Application/components/Search/Autocomplete/SearchResults/ClientDob.js
ggomez-cambriasolutions: I know, I this came from Intake, and I didn't change it.

app/javascript/Application/components/Search/SearchAssessmentHistory.test.js
ggomez-cambriasolutions: oh ok, that would be good to have different lint rules for the unit test files.

ggomez-cambriasolutions:
jenkins retest this please

ggomez-cambriasolutions:
jenkins retest this please

app/javascript/Application/components/Search/Autocomplete/SearchResults/LanguageInfo.js
ggomez-cambriasolutions: We are not displaying a primary language flag in the new search results UI, so we no longer need this method.

app/javascript/Application/components/Search/SearchHelper.js
ggomez-cambriasolutions: I've added the other residence type codes, and updated all the unit tests as well.

PR #283, opened 2018-11-16
CANS-662 Update SUD warning message.

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-662

## Description
SUD warning text no longer appears as an alert, instead it is under “No” radio button, per the design

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentFormHeader.js
```diff
@@ -126,10 +126,10 @@ class AssessmentFormHeader extends PureComponent {
 
   renderConfidentialWarningAlertIfNeeded = () =>
     this.props.assessment.can_release_confidential_info ? null : (
-      <Alert color={'warning'}>
+      <div className={'warning-text'}>
         By selecting NO, Items 7, 48, and EC 41 (Substance Use Disorder Items) from this CANS assessment will be
```
zliuwd: It should be better to save this "long text" in a constants file, by this it also could be reused in the test.

wmitchell-cambria:
@nasar-cambria - Can you attach a screenshot of how it looks now? For primarily visual changes, that helps a lot.

app/javascript/Application/components/Assessment/AssessmentFormHeader.js
wmitchell-cambria: I moved this alert to its own component in #281. It should be the same change in the other file, but let me know if run into any difficulties.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -23,6 +23,13 @@
 //********************
 //AssessmentFormHeader
 //********************
+.authorization-warning
+  margin-left: 27rem
+  
+.warning-text   
+  color: #F22613
```
wmitchell-cambria: Is there a color variable that should be used here?

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -23,6 +23,13 @@
 //********************
 //AssessmentFormHeader
 //********************
+.authorization-warning
+  margin-left: 27rem
```
wmitchell-cambria: This margin seems specific to a given screen size. Instead, we should either be using grid positioning to position it more responsively or positioning it relative to the radio button directly.

nasar-cambria: I agree if we do that it'd be responsive however we would still need to use some custom styling here. As per the requirement we need to make sure it is positioned under radio button No.

app/javascript/Application/components/Assessment/style.sass
nasar-cambria: I can ask design team. This hex code was approved by them for warning text.

app/javascript/Application/components/Assessment/style.sass
wmitchell-cambria: Yah, custom styling is fine. I am more worried about the brittleness of the styles. Maybe something like positioning it 50% left from the grid cell would work, or (as mentioned) putting it somewhere in the HTML so that it is directly aligned with the No button.

The issue with the 27rem is that it will break not only on different screen sizes, but also will be slightly off on different browsers, if they render fonts slightly differently (which they do). In that situation, the 'Yes' option may be slightly bigger or smaller, causing small errors in alignment.

nasar-cambria: True, as mentioned I can use a grid and some custom styling. for example: 
```<Col xs={4} />
          <Col xs={8}>
            <ConfidentialityAlert
              canReleaseInformation={Boolean(this.props.assessment.can_release_confidential_info)}
            />
          </Col>
``` 
with some custom style like 1 rem should be enough.

app/javascript/Application/components/Assessment/AssessmentFormHeader.js
nasar-cambria: @zliuwd I don't see this text is used anywhere else.

PR #285, opened 2018-11-16
Patch: Fix domain and DomainProgressBar propTypes

zliuwd:
## Tests

## Types of changes

## Checklist:


denys-davydov:
Could not replicate the issue you are solving here. How to do it?

zliuwd:
This console warning could be replicate locally use follow step:
1. checkout to current master branch
2. click clients list => click any clients name =>click "NEW CANS" button
3. open inspect=> console.log  you will see two block of warning messages

these warning is caused by the new functionality force the user chose the age range fist, then show the domains. 
Before a user chose a age range the  isAssessmentUnderSix will be undefined.

But in the old code, Domain component and ProgressBar component is required bool propType for isAssessmentUnderSix.

app/javascript/Application/components/Assessment/DomainProgressBar.js
```diff
@@ -53,7 +53,11 @@ class DomainProgressBar extends React.PureComponent {
 
 DomainProgressBar.propTypes = {
   domain: PropTypes.object.isRequired,
-  isAssessmentUnderSix: PropTypes.bool.isRequired,
+  isAssessmentUnderSix: PropTypes.bool,
```
wmitchell-cambria: In looking at the DomainProgressBar more closely, this should actually still be required. Otherwise, it can end up with a list of 0 items, and that can trigger a divide by zero error.

We shouldn't be rendering a DomainProgressBar for a Domain that isn't displayed. A small change to the Domain render function can make sure that we only create a `<DomainProgressBar>` with a valid isAssessmentUnderSix value.

zliuwd: :thumbsup:

PR #286, opened 2018-11-16
Update cans-api dependency version in docker-compose files

denys-davydov:
## Description
Update cans-api dependency version in docker-compose files


denys-davydov:
Retest this please

PR #288, opened 2018-11-17
Refactoring:CANS-611 Permissions routing

zliuwd:
### Refactor Permissions routing components with LoadingBoundary
Thanks for Will's advise and this handy component created by Denys !
## Tests

## Types of changes

## Checklist:


app/javascript/Application/routes/Routes.js
```diff
@@ -4,14 +4,17 @@ import { Route, Switch } from 'react-router-dom'
...
+import PermissionRedirectBoundary from './PermissionRedirectBoundary'
 
 const page = (route, navigateTo) => <Page navigateTo={navigateTo} {...route} />
 
 const Routes = () => {
   return (
     <Switch>
       <Route exact path="/">
-        <PermissionRedirect />
+        <PermissionRedirectBoundary>
```
wmitchell-cambria: We should be careful not to let too much nesting live here in the `Routes` component. This seems OK for now. If it gets more complex, we should consider making a small component to wrap it all up.

app/javascript/Application/routes/PermissionRedirectBoundary.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import UserAccountService from '../components/common/UserAccountService'
+import LoadingBoundary from '../components/common/LoadingBoundary'
+
+const PermissionRedirectBoundary = ({ children }) => (
```
wmitchell-cambria: This is a nice component, and it feels like it could be useful elsewhere. If this were called `CurrentUserLoadingBoundary`, it could live in `/common`.

PR #289, opened 2018-11-19
CANS-712 Restricting subset of account information from perry token

zliuwd:
## Description

https://osi-cwds.atlassian.net/browse/CANS-712
## Tests

## Types of changes

## Checklist:

## Remarks
- ### First of all
If you found any functionality is impacted by this change please let me know asap, it will be very appreciated! 

- ### The expected result of this story is:
Only required values (subset) for logged in user from perry token should be exposed in network traffic.

- ### what I did
After research I believe that we just need UserAccountService.fetchCurrent() return 5 pieces of data for current CANS funtionalities, and they are : "first_name", "last_name", "county_name", "privileges", "staff_id".  Then I did some restriction and current API just return these 5 pieces.

 - ### Below is the comparison chart of the data exposed in network traffic:
screen shot 2018-11-18 at 11 46 37 pm

***

screen shot 2018-11-18 at 10 23 27 pm


wmitchell-cambria:
Can you elaborate about what is sensitive about these fields?

zliuwd:
Hi Will, just talked this with Leonid. The title of this story is a little bit confusing so I remove the word "sensitive" (even it still in Jira) . What it really concerned here is not which are the sensitive fields.  The problem before this fixing is that the AccountService.fetchCurrent() will return the whole data of Perry token, this data exposed too much information of the user, in some case it could be not secured. And this data still growing, so we expect to restrict what it will expose just as what we need currently .

PR #291, opened 2018-11-20
Cans 657 capture new relic metrics

AK56:
## Tests

## Types of changes

## Checklist:

Bringing the business value of logging Analytics for CANS to life. Please read through the story https://osi-cwds.atlassian.net/browse/CANS-657 to get an idea of what initially or as per the scope of the story is proposed for Analytics.

Side Note: Something is bizarre off either my local checkout of master or something else, as it is attaching the commit history of my previous work. The only commits I did for this PR as I'm editing this are: [7949e60](https://github.com/ca-cwds/cans/pull/291/commits/7949e6052c1e120ff8ac023f887d3951894a14ea) and [5d47920](https://github.com/ca-cwds/cans/pull/291/commits/5d4792051b99dd7675cf1eabe5e75005912679ff)
I literally deleted the initial branch and closed PR-290, which I created for this story initially because it was pulling in all those commits. But doing a fresh branch of my master still logged in all those previous commits. Happy to discuss more to hunt down the issue.


app/javascript/Application/util/analytics.js
```diff
@@ -0,0 +1,21 @@
+export const logAnalytics = (actionName, attributes) => {
+  if (!window.newrelic || !actionName) {
+    return
+  }
+
+  if (attributes) {
```
denys-davydov: Please consider to change the if/else block with one line ```window.newrelic.addPageAction(actionName, attributes)```.

app/javascript/Application/util/analytics.js
```diff
@@ -0,0 +1,21 @@
...
+    window.newrelic.addPageAction(actionName, attributes)
+  } else {
+    window.newrelic.addPageAction(actionName)
+  }
+}
+export const setNewRelicAttributes = attributes => {
+  if (!window.newrelic) {
+    return
+  }
+  if (attributes) {
```
denys-davydov: To make code a little cleaner you may remove the if block above and change this line to ```if (window.newrelic && attributes) {```

app/javascript/Application/components/Layout/Page.js
```diff
@@ -22,12 +24,28 @@ class Page extends Component {
...
+    }
   }
 
   async componentDidUpdate(prevProps) {
     if (prevProps === this.props) return
     const client = await this.fetchClientIfNeeded()
     await this.setState({ client })
+    if (['CLIENT_SEARCH', 'CHILD_LIST', 'STAFF_LIST'].includes(this.props.navigateTo)) {
+      const currentUser = await UserAccountService.fetchCurrent()
+      setNewRelicAttributes({
```
wmitchell-cambria: I have a bad feeling about this pattern. I think it's that React attempts to make data flow explicit and as declarative as possible, and this function sets global state outside of the lifecycle. It is kinda like a context setting.

I would worry a little about race conditions here, since UserAccountService is asynchronous and this happens after update/mount. That means that children could be rendered and log actions before this gets set.

An alternative would be to explicitly pass the needed information to the children that log the actions. That way, they have it when they log the action, and there is less usage of global state. That may be a little more hassle than it is worth at the moment, though. But...we should test whether we are actually getting the information we need in all cases, even immediately logged actions.

If we do stick with this approach, what if this were separated out into its own component? Page is already focused on two responsibilities (layout and navigation), and this adds a third. If instead, it rendered a child component named something like `<SetAnalyticsRoute location={this.props.navigationTo} />`, that component could have the single responsibility of updating the global NewRelic state. At least in that scenario, the global state usage is isolated.

AK56: Another idea to avoid any Race conditions issue, if we capture the currentUser under <Page> state on component mount (which should happen only once) and then pretty much utilize it on update. 
`const currentUser = await UserAccountService.fetchCurrent()`
`await this.setState({ client, isLoaded: true, currentUser: currentUser })`
As much as I'm fan of modularity, I believe this is one line javascript and we are not rendering any elements, and the function 'setNewRelicAttribute' is already defined in its own file, so doing a separate component could be a little overkill

app/javascript/Application/util/analytics.js
```diff
@@ -0,0 +1,21 @@
+export const logAnalytics = (actionName, attributes) => {
```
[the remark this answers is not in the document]
AK56: Agree with the bad naming choice here.

app/javascript/Application/components/Layout/Page.js
wmitchell-cambria: Your call. The race condition could apply to mounting as well, fwiw.

app/javascript/Application/util/analytics.js
AK56: Using this pattern to bother not executing if no newrelic object present or trying to log an event without an action.

AK56:
Refactoring and renaming functions to fit in more with their functionality context. Replacing the New Relic metrics call with addPageAction for dashboard visits, but I’m keeping the code for setNewRelicAttributes as it might be handy in the near future as we potentially will be indulging into more Metrics capture. 

Please retest.

app/javascript/Application/components/Layout/Page.test.js
```diff
@@ -105,10 +116,56 @@ describe('<Page />', () => {
...
+    it('renders <SupervisorDashboard /> when navigated to', async () => {
+      const wrapper = shallow(<Page match={{ params: {} }} location={{}} navigateTo={navigation.STAFF_LIST} />)
+      await wrapper.instance().componentDidMount()
+      expect(wrapper.find(SupervisorDashboard).exists()).toBe(true)
+    })
+
+    it('calls method to log dashboard visits to newrelic', async () => {
+      const wrapper = shallow(<Page match={{ params: {} }} location={{}} navigateTo={navigation.STAFF_LIST} />)
+      const dashboardNewRelicSpy = jest.fn()
+      wrapper.instance().logDashboardVisitToNewRelic = dashboardNewRelicSpy
```
wmitchell-cambria: We should spy on the analytics module rather than the instance method. Spying on the instance method depends more on the internal implementation of the component than on the effect that the method has.

app/javascript/Application/components/Layout/Page.test.js
```diff
@@ -105,10 +116,56 @@ describe('<Page />', () => {
...
+    it('renders <CurrentUserCaseLoadPage /> when navigated to', async () => {
+      const wrapper = shallow(<Page match={{ params: {} }} location={{}} navigateTo={navigation.CHILD_LIST} />)
+      await wrapper.instance().componentDidMount()
+      expect(wrapper.find(CurrentUserCaseLoadPage).exists()).toBe(true)
+    })
+
+    it('calls method to log dashboard visits to newrelic', async () => {
+      const wrapper = shallow(<Page match={{ params: {} }} location={{}} navigateTo={navigation.CHILD_LIST} />)
+      const dashboardNewRelicSpy = jest.fn()
+      wrapper.instance().logDashboardVisitToNewRelic = dashboardNewRelicSpy
```
wmitchell-cambria: Same here.

app/javascript/Application/components/Layout/Page.test.js
```diff
@@ -118,13 +175,31 @@ describe('<Page />', () => {
...
+      jest.spyOn(ClientService, 'fetch').mockReturnValue(Promise.resolve(childInfoJson))
+      const wrapper = getWrapper(navigation.CLIENT_SEARCH)
+      await wrapper.instance().componentDidMount()
+      expect(wrapper.find(SearchContainer).length).toBe(1)
+    })
+
+    it('calls method to log dashboard visits to newrelic', async () => {
+      const wrapper = getWrapper(navigation.CLIENT_SEARCH)
+      const dashboardNewRelicSpy = jest.fn()
+      wrapper.instance().logDashboardVisitToNewRelic = dashboardNewRelicSpy
```
wmitchell-cambria: And here.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -242,6 +243,13 @@ class AssessmentContainer extends Component {
         this.setState({ assessmentServiceStatus: LoadingState.error })
       }
     }
+    if (this.state.assessment.id) {
+      // Capture New Relic data after the assessment has been successfully saved
+      logPageAction('assessmentsInProgress', {
```
wmitchell-cambria: Actions are more clear if they are named as verbs, rather than nouns. For example, `assessmentSave` makes it clear that the user saved the assessment. `assessmentsInProgress` describes the state of the assessment, but not the "page action" that the user took. It's minor now, but keeping consistent naming patterns will make it easier to manage New Relic metrics down the line.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -277,6 +285,13 @@ class AssessmentContainer extends Component {
         this.setState({ assessmentServiceStatus: LoadingState.error })
       }
     }
+    if (this.state.assessment.id) {
+      // Capture New Relic data after the assessment has been successfully submitted
+      logPageAction('assessmentsCompleted', {
```
wmitchell-cambria: Similar to previous comment, naming this `assessmentSubmit` would be clear about the user action, to the point where the comment for additional explanation is unnecessary.

app/javascript/Application/components/Layout/Page.test.js
wmitchell-cambria: Ah, I see. In that case, since the `logPageAction` behavior is tested elsewhere, this test is unnecessary. It only serves to test the implementation details, which is brittle to refactoring. If the method is renamed, combined, or split, the tests will need to be updated, even though the external behavior of the component has not changed.

AK56: Agree, the mind being driven by TDD wanted to capture all pieces. I can remove this.

wmitchell-cambria: 👍 Yeah, I totally understand. This is a case where refactoring to a DRY method would be part of the Refactor phase of TDD, so it doesn't need its own tests. As you make this DRY, you ensure that all existing tests still pass.

AK56:
jenkins retest this

AK56:
jenkins retest this please

PR #292, opened 2018-11-20
Cans 743 search button on page headers

abelayt:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/App.js
```diff
@@ -18,7 +19,16 @@ const App = () => {
         <Header />
         <Sticker>
           <div role="contentinfo" className="sticky page-header-container">
-            <PageHeader pageTitle="CANS Assessment Application" button={null}>
+            <PageHeader
+              pageTitle="CANS Assessment Application"
+              button={
+                <Link to={`/search`} id="search-link">
```
wmitchell-cambria: This could be factored out into a SearchButton component if we want to keep App.js very small. That would also help avoid 'diving' in the tests.

app/javascript/Application/App.js
```diff
@@ -18,7 +19,16 @@ const App = () => {
...
+                  <Button color="primary" className="client-search-button">
+                    <i className="fa fa-search" /> CLIENT SEARCH
```
wmitchell-cambria: nit: I tend to use natural looking text, like `Client Search`, and then style it to all-caps with CSS. I think it helps with separating presentation from content.

app/javascript/Application/styles/global/global-styles.sass
```diff
@@ -31,3 +31,8 @@ a:visited
 
 .cancel-button-darker
   color: $link-on-gray-back
+
+.client-search-button
+  text-align: right
+  margin-left: 65.0rem
```
wmitchell-cambria: Should the buttons be `float: right`, rather than having a large left margin?

denys-davydov:
@wmitchell-cambria @abelayt I think that in the scope of this user story we should add the button to all the pages and then remove/change buttons for each particular page as a part of a separate user story.

wmitchell-cambria:
I agree.

app/javascript/Application/App.js
abelayt: SearchButton component done.

app/javascript/Application/App.js
abelayt: done with text-transform.

app/javascript/Application/styles/global/global-styles.sass
abelayt: WIP

PR #315, opened 2018-11-30
CANS-208. User can enter comments at the item level

denys-davydov:
## Description
https://osi-cwds.atlassian.net/browse/CANS-208

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,66 @@
...
+          <Input
+            type={'textarea'}
+            className={`item-comment-textarea${classSuffix}`}
+            value={value}
+            placeholder={'Comment'}
+            onChange={this.handleInternalValueUpdate}
+            onFocus={this.handleOnFocus}
+            onBlur={this.handleOnBlur}
+            maxLength={250}
+          />
```
AK56: Having maxLength's value to be driven by a prop instead will make it more reusable, like for Domain level we need it to be 2500.

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,66 @@
...
+          <ItemCommentIcon isSolid={Boolean(value)} className={'item-comment-icon'} />
+          {isFocused && <span className={'item-comment-text-length'}>{value.length}/250</span>}
```
AK56: And we can use that prop for our magic number here too

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,66 @@
...
+ItemComment.propTypes = {
+  comment: PropTypes.string,
+  onChange: PropTypes.func.isRequired,
+}
+
+ItemComment.defaultProps = {
+  comment: '',
+}
+
+export default ItemComment
```
AK56: Here's the summary: The ItemComment component as the name says it all is to serve only the item level comment purpose. But for it to be shareable I propose these changes:

- Component name to make it sound more like a standard. The name 'Comment' could be right as we might utilize it in other parts of App down the road. 
- className prefix can be driven by a prop as well, where for when rendering in Item, we set className='item-comment' and for domain, it is 'domain-comment' and we keep our styles intact

denys-davydov:
@AK56 As we discussed offline, I'm creating a comment component for my user story (that is actually created in this PR) and you will re-use it in your CANS-70 user story with any updates that are needed for you. So my `ItemComment` component is not a multipurpose component as of now, it solves all the issues for my user story. All the updates you've proposed are to make the component more suitable for your story's needs, which I am not aware of while working on mine. Actually, I've made my part of the plan that we've discussed offline.

app/javascript/Application/components/Assessment/ItemComment.js
denys-davydov: Will remove the magic number as of now to make it a const instead

app/javascript/Application/components/Assessment/Item.test.js
```diff
@@ -261,13 +307,48 @@ describe('<Item />', () => {
...
+          isAssessmentUnderSix={false}
+          item={{ ...itemDefault }}
+          i18n={{ ...i18nDefault }}
+          canReleaseConfidentialInfo={false}
+          onCommentUpdate={onCommentUpdateMock}
+          onRatingUpdate={() => {}}
+          onConfidentialityUpdate={() => {}}
+        />
+      )
+      wrapper.instance().handleCommentChange('new comment')
```
wmitchell-cambria: Re: my previous comment, this test is better. The only small change I would make is to find the callback on the child, rather than calling the method on the instance. That would actually make the test right above this redundant, as well.

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Input } from 'reactstrap'
```
wmitchell-cambria: It looks like the component library re-exports reactstrap components. We could import it from there to try to consolidate our direct dependencies.

https://github.com/ca-cwds/design-system/blob/master/libs/reactstrap/src/index.js#L78

AK56:
> @AK56 As we discussed offline, I'm creating a comment component for my user story (that is actually created in this PR) and you will re-use it in your CANS-70 user story with any updates that are needed for you. So my `ItemComment` component is not a multipurpose component as of now, it solves all the issues for my user story. All the updates you've proposed are to make the component more suitable for your story's needs, which I am not aware of while working on mine. Actually, I've made my part of the plan that we've discussed offline.
@denys-davydov I remember our plan! I will apply the desired changes to leverage Domain & Item level comments with one shareable Comment component.

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,68 @@
...
+
+  handleOnFocus = () => {
+    this.setState({ isFocused: true })
+  }
+
+  handleOnBlur = () => {
+    const { value } = this.state
+    const { comment, onChange } = this.props
+    if (value !== comment) {
+      onChange(value)
```
wmitchell-cambria: Is there a difference between an undefined/null comment and an empty string? It looks like we could end up with either being saved to the assessment, depending on how the user interacts with the application. Should we make an effort to convert empty strings back to undefined, so that we only have one way of representing an empty value?

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,68 @@
...
+
+  handleInternalValueUpdate = event => {
+    this.setState({ value: event.target.value })
+  }
+
+  handleOnFocus = () => {
+    this.setState({ isFocused: true })
+  }
+
+  handleOnBlur = () => {
```
wmitchell-cambria: Sometimes Safari will fire its blur event in a weird order compared to other events. We should double check that if we are focused in a comment field and click Save/Complete, that the comment is saved correctly.

app/javascript/Application/components/Assessment/ItemComment.js
```diff
@@ -0,0 +1,68 @@
...
+    const isFolded = !isFocused && !value
+    const classSuffix = isFolded ? '-empty' : ''
+    return (
+      <div className={'item-comment-wrapper'}>
+        <div className={'item-comment-block'}>
+          <Input
+            type={'textarea'}
+            className={`item-comment-textarea${classSuffix}`}
+            value={value}
+            placeholder={'Comment'}
```
wmitchell-cambria: > The placeholder attribute lets you specify a text that appears within the <input> element's content area itself when empty. It's intended to be used to show an example input, rather than an explanation or prompt, but tends to be badly misused.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Labels_and_placeholders

For accessibility, we will need a separate label for this form control.

app/javascript/Application/components/Assessment/ItemComment.test.js
```diff
@@ -0,0 +1,145 @@
...
+    })
+  })
+
+  describe('#handleOnBlur()', () => {
+    it('should be propagated to Input onBlur prop', () => {
+      const wrapper = getWrapper()
+      expect(wrapper.find(Input).props().onBlur).toBe(wrapper.instance().handleOnBlur)
+    })
+
+    it('should set isFocused state value to false', () => {
```
wmitchell-cambria: I would combine this with the test above, to avoid testing component internals too much.

app/javascript/Application/components/Assessment/ItemComment.test.js
```diff
@@ -0,0 +1,145 @@
...
+    })
+  })
+
+  describe('#handleOnFocus()', () => {
+    it('should be propagated to Input onFocus prop', () => {
+      const wrapper = getWrapper()
+      expect(wrapper.find(Input).props().onFocus).toBe(wrapper.instance().handleOnFocus)
+    })
+
+    it('should set isFocused state value to true', () => {
```
wmitchell-cambria: Same here. Instead of `wrapper.instance().handleOnFocus()`, call `wrapper.find(Input).props().onFocus()`. Then the test above is unnecessary, and the component is a little easier to refactor, such as by renaming/changing internal methods.

app/javascript/Application/components/Assessment/ItemComment.test.js
```diff
@@ -0,0 +1,145 @@
...
+    })
+  })
+
+  describe('#handleInternalValueUpdate()', () => {
+    it('should be propagated to Input onChange prop', () => {
+      const wrapper = getWrapper()
+      expect(wrapper.find(Input).props().onChange).toBe(wrapper.instance().handleInternalValueUpdate)
+    })
+
+    it('should update the state with new value', () => {
```
wmitchell-cambria: Same here. (Not sure why I'm going in reverse order in the file...)

app/javascript/Application/components/Assessment/ItemComment.test.js
```diff
@@ -0,0 +1,145 @@
...
+        expect(wrapper.find(ItemCommentIcon).props().isSolid).toBeTruthy()
+      })
+
+      it('should be outlined when a comment is empty', () => {
+        const wrapper = getWrapper('')
+        expect(wrapper.find(ItemCommentIcon).props().isSolid).toBeFalsy()
+      })
+    })
+
+    describe('comment length indicator', () => {
```
wmitchell-cambria: One pattern that I have heard in React unit testing is "group by setup, rather than behavior." The setup here is super tiny, but I still think it can be a good example to describe the concept overall.

The idea is that React test suites are easier to reason about if you group describe blocks by shared setup, in this case `describe('when a comment is not empty')` and `describe('when a comment is empty')`. Then within those blocks, there will be tests for `it('renders a solid icon')`, `it('renders a count when focused')`, etc.

Sharing the setup in this way helps keep tests small and clear. If it becomes difficult to factor it out that way and the behavior really feels like it needs to be in its own `describe` block, that is usually a code smell that the behavior is complex enough to live in its own component that can manage just that one thing. (Example, `<LengthIndicator isFocused={isFocused} text={value} max={250} />`)

Anyways, this is a super simple case here, so I'm not sure it's worth refactoring. I'm just using it as an excuse to write down the idea.

app/javascript/Application/components/Assessment/ItemCommentIcon.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Icon } from '@cwds/components'
+
+const ItemCommentIcon = ({ isSolid, className }) => (
+  <Icon
+    name="comment-alt"
+    set={isSolid ? 'fas' : 'far'}
```
wmitchell-cambria: So we have these `far` and `fas` icons included now? They weren't available when I was working on the tool tips last week.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -228,6 +228,52 @@
...
+  justify-content: center
+
+.item-comment-block
+  display: flex
+  flex-wrap: wrap
+  position: relative
+  width: 60%
+
+.item-comment-textarea
+  padding-left: 4rem !important
```
wmitchell-cambria: What's going on here with all the `!important` markers?

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -228,6 +228,52 @@
...
+  font-size: 1.5rem !important
+  min-height: 7rem !important
+  &::placeholder
+    color: $light-gray !important
+
+.item-comment-textarea-empty
+  @extend .item-comment-textarea
+  min-height: 3.2rem !important
+  height: 3.2rem !important
```
wmitchell-cambria: There are some magic looking numbers in these styles. 3.2rem here, and 7rem above, and to some extent the `top: 1.6rem` below. These values seem to be fixed to match specific paddings and margins of the underlying text area. They are likely to break or shift if the underlying library updates styles related styles, or if the font-size changes relative to the root font size.

I don't have an easy solution for fixing it. Maybe using `em` instead of `rem` in some places, and relating values to each other using sass variables (like the padding-left of the text area and the icon's positioning).

app/javascript/Application/components/Assessment/ItemComment.js
denys-davydov: Inside of the component we are always using and empty string for the 'value' variable. This value will be sent outside only when: (1) the component loses focus and (2) the value is changed comparing to initial/previous value. So if you simply focus and blur the commponent nothing happends. 
The empty string can be saved to the back end by only one scenario: someone removes an existing comment, leaves it blank and removes focus from the component. Then saves the assessment form. Still don't see it as a big matter.

app/javascript/Application/components/Assessment/ItemComment.js
denys-davydov: I've validated the scenario in Safari browser, works as expected.

app/javascript/Application/components/Assessment/ItemComment.js
denys-davydov: Thank you for the link, it is very useful. I think it is a designops issue, first of all, but I think, that the main concern of the placeholder misusage is covered by the comment icon. So the user can understand the input field purpose at any time. However, I've made a presentation of the component to Robert and Sonoltra and they've approved the implemented approach. We still can discuss it as a sidebar after the stand up.

app/javascript/Application/components/Assessment/ItemComment.js
wmitchell-cambria: I guess that seems fine. If we really wanted to prevent empty strings and null values from both existing, we would have to do it on the API anyways.

app/javascript/Application/components/Assessment/ItemComment.js
wmitchell-cambria: I agree that it's a design issue. It is one that we need to solve, though. Having a form element without a label triggers an alert using the Wave tool:
screen shot 2018-12-03 at 3 39 22 pm

Screen readers will have a hard time figuring out that the comment icon is a label for the field, or that the icon even represents a comment.

wmitchell-cambria: Sticking with the same designs from Intake, with a label outside of the text field, will pass the accessibility tests, and Design and PO both signed off on that approach in refinement sessions, I think. That could be a good way to address the issue.

app/javascript/Application/components/Assessment/ItemComment.test.js
denys-davydov: Thank you I'll consider this approach for the future

app/javascript/Application/components/Assessment/ItemCommentIcon.js
denys-davydov: Here is the design-system PR https://github.com/ca-cwds/design-system/pull/270

app/javascript/Application/components/Assessment/style.sass
denys-davydov: When the input was imported from the reactstrap library they've been needed. Now, after using 'cwds/components' Input component, they will be gone.

app/javascript/Application/components/Assessment/ItemComment.js
denys-davydov: Used the one from our library and it helped to avoid a bunch of ```!important``` later

app/javascript/Application/components/Assessment/ItemComment.test.js
denys-davydov: Fixed

app/javascript/Application/components/Assessment/ItemComment.test.js
denys-davydov: Fixed

app/javascript/Application/components/Assessment/ItemComment.test.js
denys-davydov: Fixed

denys-davydov:
@wmitchell-cambria Please, revisit the PR

app/javascript/Application/components/Assessment/style.sass
wmitchell-cambria: That _is_ the problem. The goal is to maximize the good changes we get automatically while minimizing the damage to our overrides. In general, positioning things relative to other elements helps maintain structure as things change. For example, if the underlying font-size doubles, the text area should double, since we really care that it is 1 or 3 lines, not that it is 32px. I don't have a laundry list of good and bad styles for this, but the fewer particular numbers we have, the better. Nice round percentages, whole number increments of `em`, sass variables that relate certain spaces to each other... Those tend to be more maintainable. Odd numbers and pixel-perfect positioning, conversely, are just a little smelly.

This is not a blocking issue, though. It's just something we should keep an eye on in general. CSS can easily become a wiry mess of interconnected rules, and then we have to override everything everywhere, and we get more susceptible to small upstream changes or browser differences.

app/javascript/Application/components/Assessment/ItemCommentIcon.js
wmitchell-cambria: Great, thanks for the link!

denys-davydov:
Retest this please

denys-davydov:
@wmitchell-cambria After a discussion with Robert and Bryan we've decided to add a label and move the icon out of the textarea element, so that the ItemComment component becomes accessible. 
accessible-comment

denys-davydov:
Retest this please

PR #316, opened 2018-11-30
Move Print button in Change Log to header;

ggomez-cambriasolutions:
## Description
Move Change Log Print Button to Page Header using Page Header Buttons Builder.

## Tests

## Types of changes

## Checklist:


ggomez-cambriasolutions:
jenkins retest this please

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
```diff
@@ -30,27 +31,42 @@ const columnConfig = [
...
+  }
+
+  render() {
+    const { client, changeHistory } = this.props
+    const changeHistoryLength = changeHistory.length
+    const minRows = 0
+    const defaultPageSize = 10
+
+    const clientDob = client.dob
+      ? moment(client.dob)
```
denys-davydov: Use ```dateHelper.isoToLocalDate()``` function to format dates

ggomez-cambriasolutions: ok sound good @denys-davydov

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
```diff
@@ -30,27 +31,38 @@ const columnConfig = [
...
-    <Fragment>
-      <Row>
-        <Col xs="12">
-          <PrintButton
-            className={'change-log-print-button'}
-            node={<PrintChangeLog history={changeHistory} client={client} assessmentId={match.params.id} />}
-          />
-        </Col>
-      </Row>
+    this.initHeaderButtons(enablePrintButton)
```
wmitchell-cambria: Heads up that if this component updates, it will not re-render its header buttons. Here are two imaginary (but plausible) situations where that may occur:

1. We change the ChangeLogBoundary to render the ChangeLog while it is still loading. Instead of rendering `null` on an empty history, we render a loading spinner. In that case, the print button will be initialized to always printing the empty history.
2. We add a `Refresh` button to the page, which gets the latest changes since you started viewing it. The initial history is already locked into the Print button, and further updates will not be printed until you navigate away or hard refresh the page.

In those scenarios, with those seemingly unrelated changes, how confident are we that we will catch a bug in this behavior? Perhaps we should add an `componentDidUpdate` function now.

PR #322, opened 2018-12-04
Checking in base New Relic metrics for CANS

AK56:
Documentation for our New Relic Metrics for CANS. This gives us ground to build dashboards across different environments for CANS and also to fallback on if the end user manipulation with New Relic messes up the metrics tables/charts.


AK56:
@wmitchell-cambria Good point! We can create a folder 'resources' and a docs folder within that to hold documentation. The resources folder can also hold any other support files/folders.

denys-davydov:
@AK56 What if we create a *newrelic* folder in the root folder of our project and store the file there? It will add visibility to the file. It also could be a well-formatted *readme.md* file for convenience sake.

AK56:
@denys-davydov I feel like we should have a parent folder to hold all application support stuff including documentation and any other resources we might need to put in there. And considering your idea of readme naming, how about we put this documentation under resources/docs/new-relic/readme.md ?

denys-davydov:
@AK56 @wmitchell-cambria I'm fine with any folder structure unless it is served to the end user. Just not fond of hiding the file behind ten folders when it is the only file we have. Trying to keep it simple 😎

wmitchell-cambria:
I agree completely. My sarcasm may not have been strong enough.

denys-davydov:
@wmitchell-cambria Sarcasm often loses power when transmitted over the internet 😿

wmitchell-cambria:
https://en.wikipedia.org/wiki/Poe%27s_law

PR #326, opened 2018-12-04
CANS-797 Remove Sidenav from dashboard

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-797

## Description
Remove sidenav from dashboard and acceptance test fix.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Layout/Page.js
```diff
@@ -163,6 +129,13 @@ class Page extends Component {
             url={this.props.match.url}
           />
           {this.renderRow()}
+          <Row>
```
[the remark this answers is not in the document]
LeonidMarushevskyi: +1 to simplifying this construction.

spec/acceptance/assessment_page_spec.rb
```diff
@@ -5,7 +5,7 @@
 feature 'Assessment Page' do
   scenario 'can fill and complete assessment' do
     login
-    click_link 'County Client List'
+    visit '/clients'
```
wmitchell-cambria: Will this ever fail? It is followed immediately by another `visit`, and it matches the page the user lands on anyways.

app/javascript/Application/components/Layout/Page.js
```diff
@@ -105,44 +103,12 @@ class Page extends Component {
...
-    )
-  }
-
   renderRow() {
     switch (this.props.navigateTo) {
       case navigation.CLIENT_SEARCH:
       case navigation.ASSESSMENT_CHANGELOG:
       case navigation.ASSESSMENT_ADD:
       case navigation.ASSESSMENT_EDIT:
-        return this.renderWithoutSidebar(this.renderContent())
```
wmitchell-cambria: `renderRow` does nothing. Cleanup!

spec/acceptance/index_page_spec.rb
```diff
@@ -14,12 +14,14 @@
   scenario 'Supervisor lands on staff list and can logout' do
     login supervisor_json
     expect(page).to have_content('CANS')
+    sleep 1
```
wmitchell-cambria: We can probably use a `wait` option on the following expectation, which will be more clear than a `sleep`.

app/javascript/Application/components/Layout/Page.js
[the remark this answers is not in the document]
nasar-cambria: ok, make sense

spec/acceptance/index_page_spec.rb
LeonidMarushevskyi: Why we need this sleep? Isn't Copybara will wait for 10 sec according to Capybara.default_max_wait_time?

nasar-cambria: default_max_wait_time is when it's redirecting from one page to another.

spec/acceptance/assessment_page_spec.rb
nasar-cambria: it shouldn't the reason I am doing it twice is to ensure it picks up the url on a second instance if it skips on the first.

wmitchell-cambria: When would it skip on the first?

spec/acceptance/index_page_spec.rb
wmitchell-cambria: That isn't what I am suggesting. You should be able to add a wait option to the expectation directly:

`expect(page).to have_content('Assigned Staff', wait: 11)`

wmitchell-cambria: No, I think Leonid is right. `default_max_wait_time` should apply to any expectation that waits. https://github.com/teamcapybara/capybara/blob/master/README.md#asynchronous-javascript-ajax-and-friends

spec/acceptance/assessment_page_spec.rb
nasar-cambria: it won't. It's just my assumption. I am using it twice becz of slow render. So, to ensure it gets redirected to ```/clients/Clinet_id``` I am using another ```visit```

wmitchell-cambria: OK. Seems to me like it will just make the whole test that much slower, but your call.

spec/acceptance/index_page_spec.rb
nasar-cambria: oh ok, I tried that and it fails. 

```Failure/Error: expect(page).to have_content('Assigned Staff', wait: 10)
       expected to find text "Assigned Staff" in "CWDS\nCANS Assessment Application```

Comment (nasar-cambria) [reply spec/acceptance/index_page_spec.rb]: ok, it does say that it waits for any ajax calls and it waits on the elements to appear but here the problem is we are simply expecting something as soon as we log in. There is no interaction after the login and it fails every time.

Comment (wmitchell-cambria) [reply spec/acceptance/index_page_spec.rb]: `sleep 1` should be pretty much equivalent to adding an extra second onto the `wait` time. I'm not sure why one would fail and not the other. What are you setting the wait time at when you are testing?

Comment (nasar-cambria) [reply spec/acceptance/index_page_spec.rb]: same as suggested 
```
expect(page).to have_content('Assigned Staff', wait: 11)
```

PR #329, opened 2018-12-05
CANS-796 - LoadingBoundary "updates"

wmitchell-cambria:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-796

## Description

This is an update to LoadingBoundaries loosely related to CANS-796. For that refactoring spike, I needed a LoadingBoundary that would not just fetch data once, but re-fetch data each time its parameters changed. While the rest of the refactoring is not yet ready, this is a useful util that I think can be merged to master.

An example use case is that of search. When the query is updated, we want to fetch a new set of results. Other examples could include a Supervisor viewing different case loads by Staff ID. Basically, any time the params to the LoadingBoundary change, we want it to re-fetch data based on those params.

At the moment, the LoadingBoundary preserves the previous data while it is updating. Options can be added in the future to hide the children if necessary.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/LoadingBoundary.js
```diff
@@ -12,22 +12,46 @@ class LoadingBoundary extends React.PureComponent {
   }
 
   async componentDidMount() {
+    this._mounted = true
```
wmitchell-cambria: As mentioned in the commit message, I am not happy with needing this property. However, unless we introduce a cancellable Promise library, we need it to avoid setting state after the component unmounts, which causes a warning.

wmitchell-cambria:
Closing this, as I think there is an issue with extra rendering cycles causing excessive fetch requests.

PR #331, opened 2018-12-05
Update Rails for security patches

wmitchell-cambria:
## JIRA

No story, flagged by GitHub

## Description

This update is in response to the following security vuln patches:
https://weblog.rubyonrails.org/2018/11/27/Rails-4-2-5-0-5-1-5-2-have-been-released/

## Tests

## Types of changes

## Checklist:


denys-davydov:
Shouldn't we update the base docker image *cwds/ruby*?

wmitchell-cambria:
Good catch. Let me check with @ratneshraval, but I think we should definitely update the base image as well.

wmitchell-cambria:
Chatted with @ratneshraval. We should follow up with a patch to the base image as well. As long as the app-specific Rails version is patched, we will be secure. Build times will be a little slower until we update the base image to match, though, since it has to install the second version.

LeonidMarushevskyi:
@wmitchell-cambria will rebuild base CANS image to accommodate this changes. So build time will not be affected.

PR #336, opened 2018-12-06
Revert "CANS-605 shared redis session gem"

wmitchell-cambria:
Reverts ca-cwds/cans#328

I suspect (but have not yet confirmed) that this is the cause of the access_denied page in Preint and Integration. I'm making this PR to start the review/build/merge process just in case.


ratneshraval:
@wmitchell-cambria @LeonidMarushevskyi  Let's revert this for now.

PR #338, opened 2018-12-06
Set dev Procfile making it easy to run in dev mode

ratneshraval:
## JIRA
n/a

## Description
- Procfile to run webpack and rails with single command `yarn dev`
- You will need to install gem `foreman`

## Tests

## Types of changes

## Checklist:


Procfile.dev
```diff
@@ -0,0 +1,2 @@
+web: bundle exec rails s
+hot: ./bin/webpack-dev-server
```
wmitchell-cambria: I'm not an expert with Foreman. Do you think this should use `webpack-dashboard` like the `yarn start` command does? (I don't like the dashboard, so I just run `./bin/webpack-dev-server` directly.)

ratneshraval: This will run both webpack (watch mode) and rails.

denys-davydov:
@ratneshraval Is a developer supposed to install **foreman** manually or you can add it to the development group in **Gemfile**?

ratneshraval:
@denys-davydov  both are possible options. In cals we have it declared in `gemfile`. I think that will be easier for everyone. I'll update.
@wmitchell-cambria I'll update readme.

ratneshraval:
Knock knock...

wmitchell-cambria:
The code looks safe to me, but since I am biased towards having fewer different ways to run the application, I'm going to wait for another developer to approve it. That way we have at least two people who would actually use it (because I would not).

ratneshraval:
retest this please

PR #339, opened 2018-12-06
CANS-772. Print comments at the domain level

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-772

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Print/PrintAssessment.js
```diff
@@ -34,6 +35,8 @@ import moment from 'moment'
 
 const isItemHidden = item => item.confidential_by_default && item.confidential
 
+const hasConfidentialItems = domain => domain.items.filter(isItemHidden).length > 0
```
wmitchell-cambria: `Array.prototype.some` and `Array.prototype.every` can be useful for functions like this.

`domain => domain.items.some(isItemhidden)`

app/javascript/Application/components/Print/PrintAssessmentStyle.js
```diff
@@ -41,6 +41,13 @@ export const domainTitleStyle = {
   fontWeight: 500,
 }
 
+export const domainComment = {
+  fontSize: '0.8rem',
+  fontStyle: 'italic',
+  backgroundColor: 'transparent',
+  pageBreakInside: 'avoid',
```
wmitchell-cambria: I wonder whether or not we should be using `pageBreakInside: avoid` here. Domain comments can be pretty long, so it could result in a lot of extra pages being printed if we avoid breaking. It's also possible that there will be natural paragraph breaks anyways.

Related question: Do domain comments print with no newlines, similar to item level comments? That may turn paragraphs into a really big blobs that could be hard to read.

denys-davydov: Here is an example of the long domain comment https://osi-cwds.atlassian.net/secure/attachment/19213/19213_long_domain_comment_example.pdf
We can discuss offline if any improvement is needed since the user story has nothing to say about the formatting

PR #340, opened 2018-12-06
CANS-789 Disable Complete button if user don't have permissions to complete assessment

vladimirPe:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-789

## Description
Implementation and tests

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/AuthBoundary.js
```diff
@@ -4,16 +4,18 @@ import LoadingBoundary from './LoadingBoundary'
 import SecurityService from './Security.service'
 
 class AuthBoundary extends React.PureComponent {
-  disabled = permission => {
-    return SecurityService.checkPermission(permission).then(isAuthorized => !isAuthorized)
+  disabled = (permission, andCondition, orCondition) => {
```
denys-davydov: A good idea is to call your function with a verb, not a noun

app/javascript/Application/components/common/AuthBoundary.js
```diff
@@ -23,13 +26,24 @@ class AuthBoundary extends React.PureComponent {
...
+AuthBoundary.defaultProps = {
+  andCondition: true,
+  orCondition: false,
+}
+
 const buildCreateAssessmentPermission = clientIdentifier => {
   return `client:createAssessment:${clientIdentifier}`
 }
 
+const buildCompleteAssessmentPermission = clientIdentifier => {
```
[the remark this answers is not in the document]
vladimirPe: done

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -351,7 +351,7 @@ class AssessmentContainer extends Component {
     }
     const canPerformUpdates = isReadyForAction(assessmentServiceStatus)
     const isUnderSix = assessment && assessment.state && assessment.state.under_six
-
+    assessment.person = client
```
denys-davydov: A good idea to set person in the assessment is when the assessment is fetched successfully and is going to be saved in state. ```render()``` function is called multiple times before and after that so it doesn't seem to be a good place for this operation.

Note that the assessment is fetched in 2 places with different calls.

app/javascript/Application/components/common/AuthBoundary.js
```diff
@@ -23,13 +26,30 @@ class AuthBoundary extends React.PureComponent {
...
+AuthBoundary.defaultProps = {
+  andCondition: true,
+  orCondition: false,
+}
+
 const buildCreateAssessmentPermission = clientIdentifier => {
   return `client:createAssessment:${clientIdentifier}`
 }
 
+const buildCompleteAssessmentPermission = assessment => {
```
denys-davydov: I would suggest moving ```buildCreateAssessmentPermission()``` and ```buildCreateAssessmentPermission()``` functions to a helper class.

wmitchell-cambria:
Is this still a work-in-progress, or is it ready to review? The PR title still says WIP.

vladimirPe:
@wmitchell-cambria  it is ready for review (backend part is in-progress)

app/javascript/Application/components/Assessment/AssessmentContainer.js
wmitchell-cambria: In addition to Deny's point about render being called multiple times, this also modifies state without calling `setState`, which is a big no-no in React. Mutating state like this will not cause child components to render, which leads to them being out of sync with their surrounding components.

LeonidMarushevskyi:
@wmitchell-cambria The key point of this approach is that authorization logic is transparent for frontend application and implemented on the backend. 

**Pros**
* No hardcoded authorization logic in JavaScript. Instead, API is used to authorize operations. In such way, the application becomes dynamic and can be configured using IDM. * The same authorization rules are used for frontend and backend authorization. No duplication of logic (we had experience doing the same different way in RFA and faced a lot of issues trying to maintain the same logic in two places). 
* Authorization logic (read, write, update, delete, etc.) is implemented in a form of decision table which can be easily reviewed, supported and modified by business people. Current decision table https://osi-cwds.atlassian.net/secure/attachment/18886/AccessDecisionTable001.xlsx The same decision table is used by Drools engine. 

**Cons**
* You can suggest some

**Regarding scalability**
1. I don't think it is a problem. It is always a goal for frontend application to minimize the number of API calls. The security API call is no different from any other API call. 
2. If we will have a performance issue, we can use different caching strategies like simple static cache inside AuthBoundary in JavaScript to make sure that we are not invoking Security API more then we need. Also, we can come up with a more sophisticated cache on the Rails (API Gateway) side. Backend API can be scaled if needed and some parts of logic cached to achieve acceptable performance.

wmitchell-cambria:
I think those Pros make a lot of sense, thanks.

Intake is an interesting case, really. Those endpoints you linked actually just read from the user account info in the Rails session. Those particular endpoints would be better off replaced by reading the user account we already have stored in JS, rather than making a new API call. We use those endpoints as partial checks to see if the user _definitely does not have access_ to perform a certain API call (in this case, adding a client to a Snapshot from search). If we can't tell, because the user has an override privilege, the client has multiple counties, etc, we go ahead and make the API call, and handle the potential error response gracefully.

I am not sure I would call that an example of this pattern, where all of the logic is handled in the back-end. We intentionally duplicate the logic in the front-end, because authorization happens right when the user takes an action and we cannot afford to wait (otherwise, we would just make the real API call). That being said, we have had a few small issues with logic drift, where the front-end was preventing a case that the back-end was allowing. An interesting note is that case, which included clients in multiple counties, was not defined in our authorization matrix. Snapshot is an example of a case where this approach will not work from a user experience, and it also illustrates that the authorization matrices can be incomplete no matter how we implement it.

Anyways, this case is a little different, since it does happen before the actual point of user interaction, so we have the time to make a pre-emptive API call if necessary. In this situation, I think having the API handle the logic does make sense. The current approach suffers from a few cons, that I can see:

* We have to wait for the API call. It would be simpler if we had the information as soon as the surrounding components rendered. With an asynchronous call, there are questions of what to do while it is in-flight or if it fails. Plus, in some situations such as Search, we need the information right when the user performs an action for it to be useful. (In this way, the security API call _is_ different from other API calls.)
* We will be making a lot of these API calls, and potentially making them multiple times. As you mention, we could handle this with caching at various levels. I treat this as a con compared to an approach that would not require caching.

Here's a potentially crazy idea... What if instead of having a Security Check endpoint, we included these in the responses for various related resources? For example, when you fetch an Assessment, it would include all of the actions available to perform on that Assessment, and that list of actions could be filtered by what the user has permission to do. I got this idea from the [Richardson Maturity Model for REST](https://martinfowler.com/articles/richardsonMaturityModel.html#level3). It would offer the following benefits:

* The available actions would be immediately accessible to the front-end as soon as we fetched the data, so we would not have to deal with asynchronous loading of permissions.
* We would not need to make a separate API call at all, so we would not need to make separate calls for Edit and Delete on the same resource. Also, no need for caching in JS or Rails.
* Also due to the above, it would support use cases like Search where we need immediate access for it to be useful.
* The permissions would be delivered with their relevant resources, which means they would be naturally managed in the same location in React, without needing to construct a special permission string.
* The API benefits by becoming more RESTful. +1 Self Discovery and +1 HATEOAS.

Cons:

* Slightly more data included in API responses. (Useful data, though!)
* If the computation is complex, then computing it on each API call may be slow. (But as you mentioned, API can scale and we could probably cache it there as well.)
* It's different. (But so is this other new way.)

Thoughts? Maybe this is complex enough to discuss at the Senior Dev meeting tomorrow.

_Side note: I'm trying to focus on the overall approach before focusing on the JavaScript itself._

wmitchell-cambria:
Chatted with Leonid offline. We're going to see about switching to a similar approach to the one I described, but for this story this sprint, it seems reasonable to move forward with the `AuthBoundary` approach to get something to done.

@vladimirPe - Can you Slack me once you push the changes you were working on, and I'll review the JS in the morning?

wmitchell-cambria:
👍  Reviewing this PR this morning.

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,32 +1,30 @@
...
-        <Button
-          id={'submit-assessment'}
-          className={'button-fix-primary'}
-          color={'primary'}
-          disabled={!isSubmitButtonEnabled}
-          onClick={onSubmitAssessment}
-        >
-          Complete
-        </Button>
+        <AuthBoundary permission={buildCompleteAssessmentPermission(assessment)} andCondition={isSubmitButtonEnabled}>
```
wmitchell-cambria: What happens if the permission is null? We should also consider what happens while the API call is in flight or when it has an error response. Right now, it looks like it enables the button, which I think makes sense. (If the user is fast or the preliminary check fails, the API will just catch them when they try to submit.)

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,32 +1,30 @@
...
+          <CompleteAssessmentButton onSubmitAssessment={onSubmitAssessment} />
+        </AuthBoundary>
       </div>
     )
   }
 }
 
 AssessmentFormFooter.propTypes = {
+  assessment: PropTypes.object.isRequired,
```
wmitchell-cambria: We should define some shape here. We do not care about the entire assessment, but it needs at least the properties that `buildCompleteAssessmentPermission` expects. (Alternatively, we could pass just the individual props needed to calculate that assessment, so that we do not re-render unless those few properties change.)

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -178,6 +178,7 @@ class AssessmentContainer extends Component {
 
   updateAssessment = assessment => {
     const isValidForSubmit = validateAssessmentForSubmit(assessment)
+    assessment.person = this.props.client
```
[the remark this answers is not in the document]
vladimirPe: There are several places where a person is set to assessment so my assumption is that assessment can be without a person in the updateAssessment function and the updateAssessment function is the only place where setState called with assessment.

app/javascript/Application/components/common/AuthBoundary.js
```diff
@@ -4,17 +4,26 @@ import LoadingBoundary from './LoadingBoundary'
 import SecurityService from './Security.service'
 
 class AuthBoundary extends React.PureComponent {
-  disabled = permission => {
+  isDisabled = permission => {
+    if (!this.props.andCondition) {
```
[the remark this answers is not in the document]
vladimirPe: AuthBoundary is designed to disable the children components if check permission returns false. `andCondition` and `orCondition ` is an extension to support external logic.
In case of hasPermission prop we need to rework or create wrappers for all the children components of AuthBoundary to support `hasPermision` prop.  In this case, why would we need AuthBoundary at all? SecurityService.checkPermission should be enough.

app/javascript/Application/components/Assessment/CompleteAssessmentButton.js
```diff
@@ -0,0 +1,26 @@
...
+    color={'primary'}
+    onClick={onSubmitAssessment}
+    disabled={disabled}
+  >
+    Complete
+  </Button>
+)
+
+CompleteAssessmentButton.propTypes = {
+  disabled: PropTypes.oneOfType([PropTypes.bool]),
```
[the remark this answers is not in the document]
wmitchell-cambria: (nonblocking) Instead of changing the propType to avoid the Lint rule, we should use a comment to disable and re-enable it. Lint should help us write good code. If we are writing workarounds just to get around Lint, it is not doing its job.

_*grumbles about how the suggested `hasPermission` prop does pass the lint rule_*

app/javascript/Application/components/common/AuthBoundary.js
```diff
@@ -1,35 +1,61 @@
...
-      </LoadingBoundary>
-    )
+    const { children } = this.props
+    const { loadingState } = this.state
+    return LoadingState.waiting === loadingState ? null : React.cloneElement(children, { disabled: this.isDisabled() })
   }
 }
 
 AuthBoundary.propTypes = {
+  andCondition: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
```
wmitchell-cambria: It does not look like these should actually be functions. I am guessing this is another case where we should disable/re-enable the boolean prop Lint rule instead?

This is a case where the naming of the properties does not have to match any special field in the children, like `disabled` does. Is there a name that makes sense and also matches the Lint format? Like `canBeEnabled` and `isAlwaysEnabled` for `and` and `or`?

app/javascript/Application/components/common/LoadingBoundary.js
```diff
@@ -49,6 +49,7 @@ LoadingBoundary.propTypes = {
 
 LoadingBoundary.defaultProps = {
   isHiddenWhileLoading: true,
+  eagerRefreshFlagObject: null,
```
wmitchell-cambria: This can be removed.

LeonidMarushevskyi:
Jenkins retest this please

app/javascript/Application/components/common/AuthBoundary.js
vladimirPe: good point, going to refactor it

PR #341, opened 2018-12-07
CANS-671 Search results only include clients under age 22

ggomez-cambriasolutions:
## JIRA
CANS-671 Search results only include clients under age 22

## Description
As a non-case carrying worker, I want to see only relevant clients in the search results so i do not have to scroll and filter useless data.

## Tests

## Types of changes

## Checklist:


app/controllers/api/people_searches_controller.rb
```diff
@@ -13,7 +13,18 @@ def index
...
+      demographic_search_params.permit(:search_term,
+                                       :is_client_only,
+                                       search_after: [],
+                                       person: {})
+    end
+
+    def demographic_search_params
+      hash = params.permit(:search_term, :is_client_only, search_after: []).to_h
+                   .merge(person: {
+                            date_of_birth: { gte: 'now-22y', lte: 'now' }
```
denys-davydov: Is ```lte: 'now'``` part required here or or it can be omitted? If it is not required I would prefer not to have it then.

ggomez-cambriasolutions: I removed the ```lte: 'now'``` statement, and it still returns client results under 22. I'm going to remove it.

PR #342, opened 2018-12-07
Cans-70- User can input comments at the domain level

AK56:
## Description
Adding comments for individual domain on an Assessment. Each domain now has an accordion appended at the bottom of its item to hold the comment. Also, there are comment icons tied to domain header, domain comment accordion header and by the comment text area. The solid color on comment icon reflects there is some input in the respective domain comment box. See the attached screenshot for details.

cans-domain_commentinaction

Dev Note: I also tweaked little logic behind the above_six_id check on the Item. Before it was done in the Item component and even though the item with undefined above_six_id wouldn't render on an above_six Assessment, the `<Divider />` component was still being rendered, messing up our styles, which became more notable after I added the domain comment accordion. Moved that logic under `<DomainListItem>` comp.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -119,6 +119,20 @@ class Assessment extends Component {
     this.props.onAssessmentUpdate(assessment)
   }
 
+  updateDomainComment = (code, comment, caregiverIndex) => {
```
wmitchell-cambria: This method uses slightly different logic than `updateItem`. Instead of using a for-of loop with breaks, which can be hard to reason about and easy to forget, we could use `domains.map` in the same way `updateItem` does. We should also make sure that the domain matching logic in the if/else block matches the logic used in `updateItem`, instead of hard-coding the 'CGV' string.

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -135,6 +135,18 @@ describe('<Assessment />', () => {
     })
   })
 
+  describe('#updateDomainComment()', () => {
```
wmitchell-cambria: One test is not sufficient for the complexity of the conditionals, branching, and looping included in the `updateDomainComment` function. There must be cases that are missing.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -119,6 +121,7 @@ class Domain extends Component {
                 </Typography>
               </Grid>
               <Grid item xs={4} className={'domain-metric'}>
+                <CommentIcon isSolid={Boolean(domain.comment)} className={'domain-toolbar-comment-icon'} />
```
wmitchell-cambria: There are no tests for this.

app/javascript/Application/components/Assessment/DomainCommentAccordion.test.js
```diff
@@ -0,0 +1,132 @@
...
+)
+
+describe('<DomainCommentAccordion />', () => {
+  it('can expand and fold', () => {
+    const wrapper = mount({ ...domainCommentAccordionDefault })
+    const foldedText = wrapper.text()
+    expect(wrapper.find('#BEHEMO-comment-accordion-expand').hasClass('fa-chevron-right')).toBe(true)
+    expect(foldedText).toMatch(/Comment/)
+
+    wrapper.find('#BEHEMO-comment-accordion-expand').simulate('click')
```
wmitchell-cambria: Rule of thumb should be one expectation per test, or more accurately, a test should only ever fail for a single reason. (given X, when Y, expect Z)

This test goes through two of those cycles. Instead, consider splitting this second half into its own test.

app/javascript/Application/components/Assessment/DomainCommentAccordion.test.js
```diff
@@ -0,0 +1,132 @@
...
+    const expandedText = wrapper.text()
+    expect(wrapper.find('#BEHEMO-comment-accordion-expand').hasClass('fa-chevron-down')).toBe(true)
+    expect(expandedText).toMatch(/Comment/)
+  })
+
+  it('when chevron button get focus and press tab accordion will not expand', async () => {
```
wmitchell-cambria: A lot of the behavior (and JSX markup) around expanding and collapsing the accordion seems to be duplicated from the `Item` component. Have you considered refactoring to bring that behavior out into a shared component instead of duplicating the code?

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
```diff
@@ -0,0 +1,101 @@
...
+                  textAlign: 'left',
+                  marginLeft: 10,
+                }}
+              >
+                {commentTitle}
+              </Typography>
+              <CommentIcon
+                isSolid={Boolean(domain.comment)}
+                className={
+                  code === 'TRM' ? 'trm-domain-comment-accordion-comment-icon' : 'domain-comment-accordion-comment-icon'
```
wmitchell-cambria: 🙅 We should avoid hard-coding domain code strings as much as possible.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -286,6 +286,61 @@
...
+  height: 0
+
+//********************
+// DomainCommentIcon
+//********************
+.domain-toolbar-comment-icon
+  margin-right: 2rem
+
+.domain-comment-accordion-comment-icon
+  margin-right: 28.310rem
```
wmitchell-cambria: Very magic number!

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -286,6 +286,61 @@
...
+
+.trm-domain-comment-accordion-comment-icon
+  margin-right: 23.725rem
```
wmitchell-cambria: This is a combination of a magic number with a reference to a specific hard-coded domain! I think these two CSS rules are worthy of some explanation.

app/javascript/Application/components/common/Comment.test.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react'
 import { shallow } from 'enzyme'
-import ItemComment from './ItemComment'
-import ItemCommentIcon from './ItemCommentIcon'
+import Comment from '../common/Comment'
+import CommentIcon from '../common/CommentIcon'
 
-describe('<ItemComment />', () => {
+describe('<Comment />', () => {
```
wmitchell-cambria: Insufficient testing. Two new props were added, and the behavior of a third was changed. And yet, no new tests?

app/javascript/Application/components/Assessment/DomainCommentAccordion.test.js
AK56: Agree

app/javascript/Application/components/common/Comment.test.js
AK56: Seems like some work has been lost in either Git stashing not applied right and yes the eye missed some of these tests. Will incorporate these.

app/javascript/Application/components/Assessment/DomainCommentAccordion.test.js
AK56: I have considered building a shared component or reusing the Item component to honor the domain comment. Besides the expand/collapse behavior, they are very different, so I wouldn't say there is lot of duplication. But I'm in favor of having a separate comp of Expand/Collapse

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
AK56: I need to start looking more at the helpers! We have `isTraumaDomain` helper that we can apply here

wmitchell-cambria: Yes, but...do we even need that? What is the reason that it needs a different class? If it is because it happens to be the last domain, then we can check that without looking at the domain code at all.

AK56: The comment icon on the domain comment accordion for the Trauma domain needs a different margin to align with the rest of items in the Trauma domain. Because the trauma domain has Yes/No in lieu of 0..5 Rating. If we have a helper to check if it's the trauma domain, why not use that then checking if it's the last domain (which is prone to change) ?

wmitchell-cambria: Oh, if it's because of the rating list, I think there is a property on the domain to match that. (Or maybe on the item level?) Let me look at this for a sec and I'll comment back with what I find.

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -119,6 +119,20 @@ class Assessment extends Component {
...
+    const assessment = clone(this.props.assessment)
+    for (const domain of assessment.state.domains) {
+      if (caregiverIndex && code === 'CGV' && domain.code === code && domain.caregiver_index === caregiverIndex) {
```
denys-davydov: This if-else block can be replaced with a short:
```
If (domain.code === code && domain.caregiver_index === caregiverIndex) {
  domain.comment = comment
  break
}
```
This should work because of the _indefined_ value for the both of the caregiver indices.

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
denys-davydov: The design looks a little messy with this icon here, but we still need to do it. I totally agree with not using domains' codes. I can propose you to use the **rating_type** field of the first item of the domain to understand if the domain has boolean type items. 
```domain.items[0].rating_type === 'BOOLEAN' ```

File: app/javascript/Application/components/Assessment/DomainItemList.js [hunk 11]
```diff
@@ -1,14 +1,16 @@
...
 import Divider from '@material-ui/core/Divider'
 import PropTypes from 'prop-types'
 
 const DomainItemList = props => {
   return props.items.map((item, index) => {
     const { code } = item
     const itemI18n = getI18nByCode(props.i18nAll, code)
-    return (
+    const isAssessmentUnderSix = props.isAssessmentUnderSix
+    return shouldItemBeRendered(isAssessmentUnderSix, item) ? (
```
denys-davydov: 👍

app/javascript/Application/components/Assessment/Assessment.js
AK56: Thank you for pointing the default undefined values for non-caregiver domains :)

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
AK56: @denys-davydov @wmitchell-cambria Thanks for pointing out the rating_type attribute. I reached out to design ops with the concern of possibly removing the comment icon from the domain comment accordion, but it seems like they would want to keep it on there for now. See some of the last comments on the story: 
https://osi-cwds.atlassian.net/browse/CANS-70

So, I will work on fixing it in a better way

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -8,6 +8,8 @@ import Tooltip from '@material-ui/core/Tooltip'
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
 import { Input } from 'reactstrap'
 import { DomainProgressBar, DomainScore, DomainItemList, DomainCaregiverControls } from './'
+import DomainCommentAccordion from './DomainCommentAccordion.js'
```
abelayt: No need to include the .js ext. here and  in Domain.test.js

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -119,6 +119,20 @@ class Assessment extends Component {
...
+        domain.comment = comment
+        break
+      } else if (code != 'CGV' && domain.code === code) {
```
abelayt: use '!==' instead of '!='

AK56:
@wmitchell-cambria please review this

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -135,6 +135,33 @@ describe('<Assessment />', () => {
...
+      domains.forEach(domain => domain.props().onDomainCommentUpdate())
+      expect(onAssessmentUpdateMock).toHaveBeenCalledTimes(domains.length)
+    })
+
+    it('updates the domain comment when invoked', () => {
+      const mockFn = jest.fn()
+      const initialAssessment = clone(assessment)
+      initialAssessment.state.domains[0] = { ...initialAssessment.state.domains[0], code: 'AGV' }
+      const wrapper = mount(<Assessment onAssessmentUpdate={mockFn} assessment={initialAssessment} i18n={i18n} />)
+      wrapper.instance().updateDomainComment('AGV', 'a domain comment')
```
wmitchell-cambria: Should we include a test case for a caregiver domain as well?

app/javascript/Application/components/Assessment/Domain.js
wmitchell-cambria: What abelayt said ^^.

app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -101,6 +105,46 @@ describe('<Domain />', () => {
...
+
+    it('should propogate onDomainCommentUpdate prop to DomainCommentAccordion props', () => {
+      const onDomainCommentUpdateMock = jest.fn()
+      const wrapper = shallow(
+        <Domain
+          key={'1'}
+          canReleaseConfidentialInfo={true}
+          domain={{ ...domainDefault }}
+          isAssessmentUnderSix={true}
+          i18n={{ ...i18nDefault }}
```
wmitchell-cambria: I'm just curious... Is there a reason to use `i18n={{ ...i18nDefault}}` instead of `i18n={i18nDefault}`? Doing it this way clones the object.

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
```diff
@@ -0,0 +1,108 @@
...
+          {isExpanded ? this.renderDomainCommentBody(id, comment) : null}
+        </div>
+        <Divider />
+      </div>
+    )
+  }
+}
+
+DomainCommentAccordion.propTypes = {
+  domain: PropTypes.object.isRequired,
```
wmitchell-cambria: nit: Not a great PropType for this, since it needs to have a list of items, with at least one item, with the property rating_type.

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
```diff
@@ -0,0 +1,108 @@
...
+              textAlign: 'left',
+              marginLeft: 10,
+            }}
+          >
+            {commentTitle}
+          </Typography>
+          <CommentIcon
+            isSolid={Boolean(domain.comment)}
+            className={'domain-comment-accordion-comment-icon'}
+            ratingType={domain.items[0].rating_type === 'BOOLEAN' ? 'bool-rating' : 'reg-rating'}
```
wmitchell-cambria: Do we have a helper method for this? We should avoid hard-coded strings as much as possible. (Specifically referring to the 'BOOLEAN', not the class names.)

app/javascript/Application/components/Assessment/Item.js
```diff
@@ -195,8 +195,8 @@ class Item extends Component {
                 naValue={this.handleNaValueSetting(rating)}
               />
             ) : null}
-            <ItemCommentIcon isSolid={Boolean(item.comment)} className={'item-toolbar-comment-icon'} />
-            <Typography variant="title" style={{ marginRight: confidential_by_default ? '4rem' : '1.5rem' }}>
+            <CommentIcon isSolid={Boolean(item.comment)} className={'item-toolbar-comment-icon'} />
+            <Typography variant="title" style={{ width: '130px', marginRight: '1.5rem' }}>
```
wmitchell-cambria: Will Material allow us to set a className so that we can move the `130px` to sass? If we could do that, then we could relate this, the `250px`/`290px` and the `85px`/`135px` to each other using variables. That would make it easier to tell that they are related to each other.

app/javascript/Application/components/Assessment/Domain.test.js
AK56: This was done with view of customizing the object. Cloning doesn't seem to hurt here though

app/javascript/Application/components/Assessment/DomainCommentAccordion.js
AK56: Good call! Not that I come across of, but worth adding one now as we are using this logic in few places.

app/javascript/Application/components/Assessment/Assessment.test.js
AK56: I thought of it including before, and since your eye caught the miss, I would definitely put it in :)

app/javascript/Application/components/Assessment/Item.js
AK56: Looked up the Typography API and Yes, we can set className onto it. Will do the same for related divs and we can then pretty much build our variables in saas to style the width.

PR #343, opened 2018-12-07
CANS-823. Receiving grey screen when clicking on Save button second time before success message disappeared

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-823

## Description
The root cause for the bug was a cloning of react component instance by using JSON.stringify() -> JSON.parse() pair. After this operation the parsed react component becomes a simple object and cannot be rendered. So instead of copying the whole array, I create a new array with the same elements in it.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/GlobalAlert.js
```diff
@@ -16,15 +15,15 @@ export class GlobalAlert extends Component {
   }
 
   onAlertEvent = ({ message, type, isAutoCloseable }) => {
-    const alerts = clone(this.state.alerts)
+    const alerts = this.state.alerts.slice()
```
wmitchell-cambria: Great catch!

If we were feeling fancy, we could avoid the clone altogether by using methods that do not mutate the underlying array. For example:

```javascript
// Push
const alerts = [...this.state.alerts, { message, type, isAutoCloseable }]
// Close
const alerts = alerts.filter((_alert, i) => index !== i)

denys-davydov: Yes, this 'clone' line looks a little excessive, but I feel it like a pattern, like I'm emphasizing that I am not changing the state directly, that I'm cloning on the very first step. IYKWIM.

PR #349, opened 2018-12-11
CANS-778 Update print version to match Assessment form

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-778

## Description
Update assessment print feature, so it prints and view accurate assessment data

When testing please save the form when you change from **6-21** to **0-5** age group because print doesn't have functionality of rendering the form when you switch the radio. You need to save the form and then click on print again to see the updated Summary Card.

**Note**: As discussed in refinement session. This story doesn't include new test cases for the changes made to print. 

## Tests

## Types of changes

## Checklist:


nasar-cambria:
jenkins re test this please.

Gemfile.lock
```diff
@@ -339,6 +342,7 @@ DEPENDENCIES
   faraday (~> 0.13.1)
   faraday_middleware (~> 0.12.2)
   feature
+  foreman
```
nasar-cambria: I just ran bundle install and these were updated.

nasar-cambria:
jenkins re-test this please

Gemfile.lock
wmitchell-cambria: They were from @ratneshraval's recent work. He marked it as an optional dependency. I am not sure if it should end up in the lock file or not.

nasar-cambria: I think I haven't ran bundle install until now after pulling in changes from master thats why these changes showed up now. Pls let me know if it's not required.

cc: @wmitchell-cambria @ratneshraval

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -69,9 +70,37 @@ class AssessmentContainer extends Component {
     this.setState({ isEditable })
   }
 
+  isSameArray = (a, b) => {
+    return a.join() === b.join()
+  }
+
+  getSummaryCode = (name, value) => {
```
wmitchell-cambria: Why is this necessary? Passing the assessment summary codes like this will cause a second render cycle of the AssessmentContainer, which is a large component with a lot of children. That will have a noticeable performance impact on the loading of this page. It also means that the data for assessment summary flows up and down the React tree.

As an alternative, consider that the Assessment Summary table is entirely dependent on the assessment and i18n codes. The PrintAssessment component already takes these as props. Just...render the AssessmentSummary as a regular old child, and use CSS if the styles need to change to fit the paper size.

On a practical note, the rendering of the assessment summary is the only part of this I have any comments on. The other small changes seem fine. I would make a separate PR just for those changes, so they can be merged while you work on the AssessmentSummary.

nasar-cambria: Done.

app/javascript/Application/components/Print/PrintAssessment.js
```diff
@@ -183,27 +185,57 @@ class PrintAssessment extends PureComponent {
...
+    const filteredDomains = domains.filter(
+      domain => (this.state.isAssessmentUnderSix ? domain.under_six : domain.above_six)
+    )
+    const summaryCodes = {
+      Strengths: this.getCodes(filteredDomains, isStrengthsDomain, item => item.rating === 0 || item.rating === 1),
+      'Action Required': this.getCodes(filteredDomains, isNeedsDomain, item => item.rating === 2),
+      'Immediate Action Required': this.getCodes(filteredDomains, isNeedsDomain, item => item.rating === imaRating),
+      Trauma: this.getCodes(domains, isTraumaDomain, item => item.rating === 1),
+    }
+    const maxObjectSize = 1
```
wmitchell-cambria: I don't really understand what `maxObjectSize` means. If we exceed the max size, we render something the summary? What is it a maximum of?

Separate from that, we probably need to filter the assessment summary from printing if the assessment is not complete. With the changes we worked on, now `summaryCodes` will always have four keys, and so `Object.keys(summaryCodes).length > 1` will always be true.

nasar-cambria: I think using max to the object size is causing confusing. My intention was to only give a variable name to avoid magic number 1 but I learned that we are not longer linting 0, 1 and 2. So,  I'll change it. 

Agreed, I'll fix that as well.

nasar-cambria:
Jenkins Retest this please.

PR #353, opened 2018-12-12
Cans 150 expand or collapse domains

abelayt:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -129,15 +136,37 @@ class Assessment extends Component {
...
     const assessmentDto = this.props.assessment
     const canReleaseConfidentialInfo = assessmentDto.can_release_confidential_info
     const assessmentJson = assessmentDto.state
     const isUnderSix = assessmentJson.under_six
     const domains = assessmentJson.domains
+    const { isDefaultExpanded } = this.state
     return (
       <Fragment>
+        {!(isUnderSix === undefined) && (
+          <Card style={{ marginBottom: 0, marginLeft: 0, marginRight: 0 }}>
```
AK56: May be we can remove these styles ?

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -20,6 +20,7 @@ const mapPropsToState = props => ({
   title: props.i18n._title_ || '',
   description: props.i18n._description_ || 'No Description',
   caregiverName: props.domain.caregiver_name || '',
+  expanded: props.isExpanded,
```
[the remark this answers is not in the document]
abelayt: I think this is an issue we need to fix not only on this but also for all affected. For now I find this to  be the simpler way to implement. Also this expansion panel with this story expand/collapse will be replaced once its equivalent is done on CWDS Component Library.

abelayt:
retest this please

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -129,26 +135,44 @@ class Assessment extends Component {
...
     const canReleaseConfidentialInfo = assessmentDto.can_release_confidential_info
     const assessmentJson = assessmentDto.state
     const isUnderSix = assessmentJson.under_six
     const domains = assessmentJson.domains
+    const { isDefaultExpanded } = this.state
     return (
       <Fragment>
+        <Card className={'assessment-expand-collapse-card'}>
+          <CardHeader className={'assessment-expand-collapse-header'}>
+            <div className={'assessment-expand-collapse-label-left'}>
```
AK56: The class name could be confusing for a div element

app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -290,6 +291,69 @@ describe('<Domain />', () => {
...
+                onAddCaregiverDomain={() => {}}
+                handleWarningShow={() => {}}
+                onCaregiverNameUpdate={() => {}}
+              />
+            )
+            domainWrapper.update()
+            expect(domainWrapper.instance().state.expanded).toBe(true)
+          })
+        })
       })
```
AK56: Do we need a test case for all domains expanded if ExpandAll after CollapseAll ?

app/javascript/Application/components/Assessment/Assessment.js
abelayt: The expand-collapse card has a header with a label: 'age range' on left and a button: 'expand/collapse' on right.  Any suggestion?

AK56: For me, `<div>` signals a wrapper or a block. Since we are rendering this inline text within 'assessment-expand-collapse-header' div, we can wrap this text in a `<span className={'assessment-expand-collapse-age-range'}>` tag

abelayt: Thanks, done.

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -129,26 +135,44 @@ class Assessment extends Component {
...
     const assessmentJson = assessmentDto.state
     const isUnderSix = assessmentJson.under_six
     const domains = assessmentJson.domains
+    const { isDefaultExpanded } = this.state
     return (
       <Fragment>
+        <Card className={'assessment-expand-collapse-card'}>
+          <CardHeader className={'assessment-expand-collapse-header'}>
+            <span className={'assessment-expand-collapse-age-range'}>
+              {isUnderSix ? 'Age Range 0-5' : 'Age Range 6-21'}
```
denys-davydov: Please add tests for this captions

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -129,26 +135,44 @@ class Assessment extends Component {
...
+            </span>
+            <Button className={'assessment-expand-collapse-button-right'} onClick={this.handleExpandAllDomains}>
+              {isDefaultExpanded ? 'Collapse All' : 'Expand All'}
```
denys-davydov: Please add tests for this captions

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -292,5 +292,18 @@ describe('<Assessment />', () => {
...
       })
     })
+
+    describe('#handleExpandAllDomains()', () => {
+      const mockFn = jest.fn()
+      const initialAssessment = clone(assessment)
+      const wrapper = shallow(<Assessment onAssessmentUpdate={mockFn} assessment={initialAssessment} i18n={i18n} />)
+      it('updates isDefaultExpanded state when Expand/Collapse Button is clicked ', () => {
+        wrapper.instance().setState({
+          isDefaultExpanded: false,
```
denys-davydov: I would add a test for a reverse case: when it is changed from _true_ to _false_

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -14,26 +14,32 @@ import { shouldDomainBeRendered } from './AssessmentHelper'
...
     this.state = {
-      ...mapPropsToState(props),
-      expanded: false,
+      expanded: props.isDefaultExpanded,
     }
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    this.setState(mapPropsToState(nextProps))
+  static getDerivedStateFromProps(nextProps, prevState) {
```
denys-davydov: This function can be replaced with a cleaner one:
```
static getDerivedStateFromProps(nextProps, prevState) {
 if (!isEmpty(nextProps.i18n) && !prevState.description) {
   return mapI18nToState(nextProps)
 }
 return null
}
```

app/javascript/Application/components/Assessment/Assessment.js
AK56: Also, I just noticed, that we are rendering this card by default on 'New CANS', means before an age-range is selected. One consequence of it is displaying 'Age Range 6-21', when no age range is selected. Thus we should make the 'assessment-expand-collapse-age-range' content to be driven by age range selected on the assessment and make 'assessment-expand-collapse-card' render after age range is selected (the way it is for Assessment Card)

denys-davydov: @AK56 The card is not rendered when _isUnderSix === undefined_, see line 154 of the file.

app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -290,6 +291,69 @@ describe('<Domain />', () => {
...
+                onDomainCommentUpdate={() => {}}
+                onRatingUpdate={() => {}}
+                onConfidentialityUpdate={() => {}}
+                onAddCaregiverDomain={() => {}}
+                handleWarningShow={() => {}}
+                onCaregiverNameUpdate={() => {}}
+              />
+            )
+            domainWrapper.update()
+            expect(domainWrapper.instance().state.expanded).toBe(false)
```
wmitchell-cambria: nit: `domainWrapper.state()` should is the same thing as `domainWrapper.instance().state`. I'm also curious why `domainWrapper.update()` is necessary. Does shallow rendering skip `getDerivedStateFromProps` by default?

app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -290,6 +291,69 @@ describe('<Domain />', () => {
...
+          })
+        })
+
+        describe('#domain panels expand when expanded is true for all domains ', () => {
+          const mockFn = jest.fn()
+          const initialAssessment = clone(assessment)
+          const assessmentWrapper = shallow(
+            <Assessment onAssessmentUpdate={mockFn} assessment={initialAssessment} i18n={i18n} />
```
wmitchell-cambria: We shouldn't need to render parents of our target component in the unit tests, except in a few rare situations (Routes come to mind). Instead, we can use the second test you have here (`it('should expand all domains by since isDefaultExpanded in the assessment is changed to true '`), and pass the isDefaultExpanded boolean directly to Domain.

The other test can be scrapped, since it has no `expect`.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -61,6 +61,31 @@ $item-confidential-label-height: 6.05rem
 .assessment-header-card-header
   border-bottom: #ccc 0.25rem solid !important
 
+.assessment-expand-collapse-card
+  margin-bottom: 0
+  margin-left: 0
+  margin-right: 0 
+
+.assessment-expand-collapse-header
+  border-bottom: #ccc 0.25rem solid !important
```
wmitchell-cambria: We should use a color variable here if possible.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -61,6 +61,31 @@ $item-confidential-label-height: 6.05rem
...
+  display: inline-block; 
+  width: 100%;
+  
+.assessment-expand-collapse-button-right
+  font-size: 1.6rem
+  color: #0e6f89
```
wmitchell-cambria: Here also. Color variable if possible.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -61,6 +61,31 @@ $item-confidential-label-height: 6.05rem
...
+  float: right
+  margin-right: 1.5rem
+  box-shadow: none
+  text-transform: none !important
```
wmitchell-cambria: I am surprised that this needs to be `!important`. I am guessing it is because we set `text-transform: uppercase` on all buttons by default, and this selector is not strong enough to override it? I wonder if we should ask DesignOps if they are _really sure_ that this button should be an exception to the main styles.

wmitchell-cambria:
@abelayt - For practical purposes, I'm going to switch my review to Approve, assuming that you will add the test for `key`. I will be OOO tomorrow, and I do not want the story to be blocked waiting for me to become available. Plus, @denys-davydov and @AK56 are still Requesting Changes, so a third seems unnecessary. :-)

app/javascript/Application/components/Assessment/Domain.test.js
abelayt: removed domainWrapper.update()  but domainWrapper.instance().state is not the same domainWrapper.state() in this case.

wmitchell-cambria: How are they different? My understanding was that `domainWrapper.state()` returns the state object for the root component being wrapped. `domainWrapper.instance()` returns that same underlying component, so I would expect them to be the same. Googling isn't helping me. :-(

abelayt:
retest this please

AK56:
We have a duplicate Card displaying age and 'Expand All' being rendered. CANS-841 got merged which is rendering the content (without the functionality of 'Expand All'). I'm attaching the screenshot, the top card is from [CANS-841](https://osi-cwds.atlassian.net/browse/CANS-841) changes in AssessmentCard.js, it doesn't have the functionality as expected but has a nice touch on the styles. 
cans-150-dupcard

The second card is from the `<DomainsHeader>` component, created here and since we have a separate component rendering the header card, we can remove the `<CardHeader>`  from AssessmentCard.js, I would suggest to co-ordinate with @DmitriyRudenko as you refactor the code, because we do want to keep the style changes from CANS-841

app/javascript/Application/components/Assessment/Domain.test.js
abelayt: Here is an example how they produce different result.
screen shot 2018-12-21 at 4 54 07 pm

abelayt: Indeed, thanks! missed the (). Sorry Will.

wmitchell-cambria: No worries! That's an easy one to miss, especially when you are expecting false or undefined.

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -129,39 +136,59 @@ class Assessment extends Component {
...
+                    canReleaseConfidentialInfo={canReleaseConfidentialInfo}
+                    onRatingUpdate={this.handleUpdateItemRating}
+                    onItemCommentUpdate={this.handleUpdateItemComment}
+                    onConfidentialityUpdate={this.handleUpdateItemConfidentiality}
+                    onAddCaregiverDomain={this.addCaregiverDomainAfter}
+                    onRemoveCaregiverDomain={this.removeCaregiverDomain}
+                    onCaregiverNameUpdate={this.updateCaregiverName}
+                    onDomainCommentUpdate={this.updateDomainComment}
+                    handleWarningShow={this.props.handleWarningShow}
+                    expandCollapse={this.expandCollapse}
```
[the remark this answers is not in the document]
abelayt: Removed along unused CardHeader import.

PR #357, opened 2018-12-13
CANS-863. Safari - Text pasted into comments isn't counted accurately with newlines

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-863

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/Comment.js
```diff
@@ -2,18 +2,35 @@ import React, { Component } from 'react'
...
     }
   }
 
   handleInternalValueUpdate = event => {
-    this.setState({ value: event.target.value })
+    const maxCommentLength = this.props.maxCommentLength
+    let value = event.target.value
+    if (value.length > maxCommentLength) {
+      value = value.slice(0, maxCommentLength)
+      event.target.value = value
```
wmitchell-cambria: Do we need to modify the event here? What happens if we skip this line?

app/javascript/Application/components/common/Comment.js
```diff
@@ -2,18 +2,35 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Label } from '@cwds/components'
 import CommentIcon from './CommentIcon'
+import { isSafari } from '../../util/common'
+
+/* Safari browser calculates the length of textarea input in a different way than other browsers.
+This function calculates how many additional symbols should be added to textarea.maxlength
+to compensate this irregular Safari behavior */
+const calculateSafariMaxLengthAddition = value => (isSafari ? (value.match(/(\r\n|\n|\r)/g) || []).length : 0)
```
AK56: Probably not too big of a deal but just an idea, if we should use isSafari boolean attribute in the state, and sets it on init calling our isSafari function. This way we call our isSafari function once in lieu of every time text area input changes.

wmitchell-cambria: `isSafari` is already a boolean. It is calculated once when the JS is loaded.

AK56: Ahh right, for some reason I was reading it as a function

wmitchell-cambria: Yeah, it's not very typical otherwise, but for browser-detection it just runs once. Easy to miss.

app/javascript/Application/components/common/Comment.js
denys-davydov: Yes, I need to explain this code block. It fixes one more bug with Safari: when textarea holds a value with *length = maxlength - 1* (e.g. maxlength is 250, the value length is 249), and you add a newline, Safari allows you to have 251 symbols. That is why I am slicing the value manually. **BUT.** The bug is not reproducible after applying the 'isSafari' workaround from this PR, so I will remove this code block.

denys-davydov:
Retest this please

PR #360, opened 2018-12-13
CANS-864 Change font icon to circle

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-864

## Description
"?" icon updated to a "i" icon

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -115,7 +115,7 @@ class Domain extends Component {
                   {title} {caregiverName && `- ${caregiverName}`}
                   {description ? (
                     <Tooltip title={description} placement="top" classes={{ tooltip: 'assessment-tooltip' }}>
-                      <i className="fa fa-question-circle domain-help-icon" />
+                      <i className="fa fa-info-circle domain-help-icon" />
```
wmitchell-cambria: Heads up that the Component Library now includes icons as well. The goal is to use those icons, which will encourage DesignOps to select from a consistent set of a handful of icons, instead of the full set. (For example, it would prevent designs from using the question icon going forward as well.)

The usage is pretty simple. One example can be found in `Ellipsis.js`. In this case, I think it would be `<Icon name="info-circle" className="domain-help-icon" />`. If you have the time to try it (like ~5 minutes, probably), it would help move us in the right direction.

nasar-cambria: Good thought but the icons from the component library looks lil small than normal size on a 100% zoom. Please see SC enclosed above. To avoid this we might have to add custom styles again. Let's, keep it as is for now and if we change it we need to change CANS summary icons as well.  
screen shot 2018-12-14 at 10 29 39 am

wmitchell-cambria: Probably a side effect of React Wood Duck setting 1rem to 10px instead of 16px. Curses!

All right, carry on. :-)

nasar-cambria:
Jenkins re test this please

PR #362, opened 2018-12-14
Patch for cans 641: fixing "add new assessment" breadcrumb bug

zliuwd:
### How to reproduce the bug
search=>history record => we will get the client's name show on the Breadcrumb, then click this name to the profile, then click the "NEW CANS" button, we will have "Client List" show on the Breadcrumb. "Client List" should not be there in this case, except user login as case-load-worker. So the staff search flow also has this problem.

### How about the fix
Adding on correct route for the "Add new assessment" and readjust the logic.

### Remarks
- Try the relative path in AddCansButton, looks works well

## This also resolve the problem that Breadcrumb border will be covered by the content border shadow when window scroll up 

https://osi-cwds.atlassian.net/browse/CANS-871 --Completed by this Pr

## Tests

## Types of changes

## Checklist:


zliuwd:
retest this please

PR #364, opened 2018-12-14
Initial commit for SemVer for PR and master

AVandana-OSI:
## JIRA
https://osi-cwds.atlassian.net/browse/FIT-349

## Description
Added 3 separate stage to make it work with SemVer, for each PR created there should be a label(MAJOR, MINOR or patch) if the label does not provide on PR Jenkins will fail at checkForLabel stage. Other 2 stages are for increment tag based on the label and other to tag the git repo. 

## Tests

## Types of changes

## Checklist:


Jenkinsfile
```diff
@@ -66,6 +72,18 @@ def checkoutStage() {
   }
 }
 
+def checkForLabel() {
+  stage('Verify SemVer Label') {
+    checkForLabel('cans')
```
wmitchell-cambria: Is it possible to import the shared functions in a way that makes it clearer which functions are imported and which are local? Most developers here are not Groovy experts, and it may be easier to debug issues if we can tell where a function is coming from. In this situation, for example, we have the overloaded `checkForLabel() // local` and `checkForLabel('cans') // shared global`. If we named the local method differently and also had an explicit import line for the functions we are using, that would be easier to trace, in my opinion.

AVandana-OSI: Sure, Good idea. I can do that

AVandana-OSI:
retest this please

LeonidMarushevskyi:
@AVandana-OSI we have couple issues after this PR. 
1. CANS Master build fails on last step Cleanup - not critical but needs to be fixed
2. Acceptance tests on Management Jenkins are failing with some error related to shared functions http://jenkins.mgmt.cwds.io:8080/job/Acceptance_tests/job/acceptance-test-cans/351/console - this is a blocker for deployment to Integration. 
Please take a look.

edgibbs:
OK, just need the shared library added to the Jenkins MGMT server should be able to get that done pretty quickly. And I'll take a look at the cleanup issue.

edgibbs:
The master build appears to be failing because of an acceptance test failure:

```
Assessment Page can fill and complete assessment
Failure/Error: expect(page).to have_content('ADD CANS', wait: 45)
```

There was one that failed in cleanup, but it looks strange anyway it was triggered like 20 something times by a pr build, probably from Adarsh's testing, if it crops up again we can look into it more, the logic didn't change for the cleanup stage.

AVandana-OSI:
@LeonidMarushevskyi  we have found the issue on cleanup and acceptance test. Once @edgibbs creates a PR on Jenkins shared library after lunch. The issue will be fixed. Right now it fails at clean up and updates manifest. But it deploys successfully to Integration. But the master build is broken because of some Acceptance Test.

PR #365, opened 2018-12-15
Cans 867 ie comment accordion alignment fix

AK56:
## Tests

## Types of changes

## Checklist:

Mostly style tweaks to make IE better understand things. The changes takes care of two issues currently being experienced in IE:

1. Domain Comment accordion header not aligning appropriately
2. Comment Icons not appearing on Domain toolbar

Bug 1 was the initial purpose for this PR but Bug 2 was found during the fix, so including the fix for it here as well. Please check out [https://osi-cwds.atlassian.net/browse/CANS-867](url) for details


app/javascript/Application/components/Assessment/style.sass
```diff
@@ -243,6 +243,9 @@ $toolbar-comment-icon-right-margin: 2rem
 .bool-rating-domain-comment-accordion-comment-icon
   margin-right: $toolbar-comment-icon-right-margin + $confidential-checkbox-width + $confidential-checkbox-right-margin + $bool-rating-width
 
+.item-confidential-block
+  $item-confidential-block-height: 6.05rem !global
```
wmitchell-cambria: This is interesting. Here, we are creating a global variable, rather than setting a style on the component. Would it be better to create the variable elsewhere, and then use it in these two locations?

Also, 6.05rem is very specific, without much indication for why it should be 60.5px. Is there some other value we could relate this to, similar to how the `margin-right` in the rule above this is calculated? That would help to identify how we are coming up with 6.05rem. Otherwise, it reads like guesswork.

AK56:
Retest this please

app/javascript/Application/components/Assessment/style.sass
AK56: Creating a global variable here is just to serve its usage down the stylesheet for 'domain-comment-accordion-comment-icon' class, and I was originally setting the height (6.05 rem) of the 'item-confidential-block' element here as well, which I believe got redacted during the tweaks. But, that answers your second question about 6.05 rem, it's being set by FormControlLabel from material UI. We can declare the variable at the top and name it to reflect that like `$confidential-label-height`

AK56:
Retest this please

AK56:
Retest this please

AK56:
Retest this please

AK56:
Retest this please

PR #390, opened 2018-12-27
CANS-902 Fixed Timeout warning to appear 5 mins prior user session expiration

AK56:
Please checkout [CANS-902](https://osi-cwds.atlassian.net/browse/CANS-902), it has deep details on what was wrong with the timeout warning not appearing 5 mins before user session expiration time.
Pretty much the culprit was `new Date().getTime()`, which was subtracted off in lieu of the 5 min warning. I have redacted it off from the TimeoutService test file as well
## Tests

## Types of changes

## Checklist:


app/javascript/Application/util/TimeoutService.test.js
```diff
@@ -39,7 +39,7 @@ describe('TimeoutService', () => {
     const timeout = new Date().getTime() + interval
     global.document.cookie = `_ca_cans_timeout=${timeout}`
     timeoutService.run()
-    const minInterval = interval - timeout - new Date().getTime()
+    const minInterval = interval - timeout
```
wmitchell-cambria: It looks like interval should be small (4 hours), and timeout should be big (now + interval). So, `minInterval` is a big negative number? Since these are unit tests, is there a way to remove `Date()` from the mix entirely, and just hardcode the times directly? Like, if `minInterval` is supposed to be 2 hours, we could just set it to 2 hours and avoid some of this math.

AK56: It's all numbers here :) We can hard code the times directly, but I lean towards still using the `Date()` in the mix, to have a timeout based on current time number and be more like our implementation. Also the `interval` and `minInterval` is going to be different for our two cases, one that calls `eventBus.post()` with interval less than 5 min and the other with higher interval. Happy to tweak them both

AK56:
retest this please

AK56:
retest this please

PR #393, opened 2018-12-28
CANS-848 - Search "Autocompleter" refactor

wmitchell-cambria:
## JIRA

[CANS-848 - Refactor Autocomplete component to use Downshift](https://osi-cwds.atlassian.net/browse/CANS-848)

## Description

This is a big PR with a lot of context. I recommend reading the story first for background. The short story is that the previous Autocompleter component copied between Intake and CANS contained a lot of tech debt. The component in Snapshot was the source of several hard-to-figure-out bugs. This refactoring story drops our dependency on `react-autocomplete` in favor of the more flexible `Downshift`.

There are some interesting components in here, including an update to our LoadingBoundary component to handle updating with new data. That change in particular requires careful attention when creating new LoadingBoundaries, as anonymous functions created during `render` can cause data to be re-fetched excessively.

Future work in this area will likely involve factoring out PersonSuggestion and InputField so that different apps can use the main logic with custom rendering.

## Tests

I tried to add some acceptance tests for these, but Capybara is pretty bad at keyboard inputs, and the way that it finds elements causes the menu to collapse. :-(

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

*Please pay attention to the LoadingBoundary changes!*

## Checklist:


wmitchell-cambria:
Now that I have a label, retest this please.

app/javascript/Application/components/Search/ClientSearch/ClientSearch.js
```diff
@@ -0,0 +1,110 @@
...
+    if (item.type === 'action') {
+      this.performAction(item)
+      this.setState({ isOpen: true })
+    } else {
+      this.props.onSelect(item)
+    }
+  }
+
+  performAction = action => {
+    if (action.action === 'Show More Results') {
```
pokharelmanish: Can we make `Show More Results` constant?

wmitchell-cambria: Good call. I think there are probably a couple other little cleanup things like this that I missed as well. I'm going through the PR now, too.

app/javascript/Application/components/Search/ClientSearch/QueryInput/QueryInput.js
```diff
@@ -0,0 +1,25 @@
...
+class QueryInput extends React.PureComponent {
+  onChange = value => {
+    this.props.onChange({ target: { value } })
+  }
+
+  render() {
+    const { onChange, ...props } = this.props
+    return (
+      <Debouncer callback={this.onChange} callbackPropName="onChange">
+        <InputField {...props} />
```
pokharelmanish: Don't we need an `id`  in the InputField?

app/javascript/Application/components/Search/ClientSearch/QueryResults/Results.js
```diff
@@ -0,0 +1,74 @@
...
+  clearItems()
+
+  const hasMoreResults = items.length < totalResults
+  const nextSearchAfter = items.length > 0 && items[items.length - 1].sort
+
+  return isOpen ? (
+    <div>
+      <div className={'suggestion-header-wrapper'} key="header" aria-live="polite">
+        <SuggestionHeader
+          page={1}
```
pokharelmanish: `1` ??

app/javascript/Application/components/Search/ClientSearch/ClientSearch.js
```diff
@@ -0,0 +1,109 @@
...
+
+  onSelect = item => {
+    if (item.type === 'action') {
+      this.performAction(item)
+      this.setState({ isOpen: true })
+    } else {
+      this.props.onSelect(item)
+    }
+  }
+
```
pokharelmanish: Shorter verion.

```
  onSelect = item => {
    if (item.type === 'action') {
      this.performAction(item)
      this.setState({ isOpen: true })
    }
    this.props.onSelect(item)
  }
```

app/javascript/Application/components/Search/ClientSearch/QueryInput/Debouncer.js
```diff
@@ -0,0 +1,34 @@
...
+
+const WAIT = 400
+
+class Debouncer extends Component {
+  constructor(props) {
+    super(props)
+    this.debounce = this.debounce.bind(this)
+  }
+
+  debounce(value) {
```
pokharelmanish: I think we can use arrow function so we don't need to bind in `constructor`

app/javascript/Application/components/Search/ClientSearch/ClientSearch.js
wmitchell-cambria: We do not want to call `onSelect` if the `if` branch is hit.

app/javascript/Application/components/Search/ClientSearch/QueryInput/Debouncer.js
wmitchell-cambria: Good catch. This was a carry-over from the previous Debouncer component.

app/javascript/Application/components/Search/ClientSearch/ClientSearch.js
pokharelmanish: oh, I thought we were returning it.

app/javascript/Application/components/Search/ClientSearch/ClientSearch.js
```diff
@@ -0,0 +1,110 @@
...
+      searchAfter: null,
+      query: {
+        searchTerm: '',
+      },
+    }
+  }
+
+  onStateChange = ({ isOpen }) => {
+    if (isOpen !== undefined) {
+      this.setState({ isOpen: isOpen })
```
ggomez-cambriasolutions: couldn't this be `this.setState({ isOpen })` instead?

wmitchell-cambria:
D'aww... thanks guys!

PR #401, opened 2019-01-03
Cans-819 Implement Front-End Assessment Delete

ggomez-cambriasolutions:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-819

## Description
Front end changes for assessment Delete functionality. Call the new endpoint, implement new logic, warning modal, & assessment card status changes.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/Ellipsis.js
```diff
@@ -23,14 +28,103 @@ export default class Ellipsis extends React.Component {
     })
   }
 
+  onItemSelect(e) {
```
pokharelmanish: we can use arrow function here so we don't need to bind in the constructor.

app/javascript/Application/components/common/Ellipsis.test.js
```diff
@@ -2,65 +2,244 @@ import React from 'react'
...
+      })
     })
 
-    describe('Link component', () => {
-      it('renders a Link component and the correct props', () => {
-        const wrapper = shallow(<Ellipsis {...defaultProps} />)
-        expect(wrapper.find(Link).props().to).toEqual({
-          pathname: '/staff/0X5/clients/C76Jg230X3/assessments/1234/changelog',
+    describe('PopOver Menu Items', () => {
+      let popOverBody
```
pokharelmanish: `popoverBody`?

app/controllers/api/assessments_controller.rb
```diff
@@ -27,5 +27,10 @@ def changes
       response = Assessments::AssessmentsRepository.new(session[:token]).changes(params[:id])
       render json: response.body, status: response.status
     end
+
+    def delete
```
pokharelmanish: Do we have a test for this?

app/javascript/Application/components/Assessment/ChangeLog/ChangeLogLoadingBoundary.js
```diff
@@ -1,19 +1,10 @@
...
-  return Promise.all([changeHistoryPromise, assessmentPromise]).then(([changeHistory, assessment]) => ({
-    changeHistory,
-    assessment,
-  }))
-}
+import AssessmentService from '../Assessment.service'
 
 const ChangeLogLoadingBoundary = ({ id, children }) => (
-  <LoadingBoundary childNodeFetchedPropName={'assessmentWithHistory'} fetch={() => fetchBoth(id)}>
+  <LoadingBoundary childNodeFetchedPropName={'assessmentHistory'} fetch={() => AssessmentService.getAllChanges(id)}>
```
wmitchell-cambria: Re: the merge conflicts in these LoadingBoundaries, my search refactors added the ability for LoadingBoundaries to update with new information. They do this when their fetch prop changes. So, passing an anonymous function like this doesn't work as intended any more, since it will create a new function (and therefore fetch new data) every single render.

In most cases, I created a 'fetchFactory' that saves a fetch in state. To solve the conflict, we just need to make the same changes to that fetchFactory that you have here.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.js
```diff
@@ -1,31 +1,37 @@
...
+    const { callback } = this.props
+    const loadingState = this.props.loadingState ? this.props.loadingState : ''
+    const noDataText =
+      loadingState === LoadingState.ready
+        ? 'No assessments currently exist for this child/youth.'
+        : 'Loading assessments...'
 
     return assessments.length === 0 ? (
-      <div id="no-data">No assessments currently exist for this child/youth.</div>
+      <div id="no-data">{noDataText}</div>
```
pokharelmanish: probably we need padding here. 
`padding-left: 1.5rem`

pokharelmanish:
retest this please

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
```diff
@@ -38,51 +38,47 @@ class AssessmentChangeLog extends Component {
...
     pageHeaderButtonsController.updateHeaderButtons(leftButton, rightButton)
   }
 
-  updatePrintButtonIfNeeded(prevChangeHistory) {
-    const { assessmentWithHistory } = this.props
-    const changeHistory = assessmentWithHistory.changeHistory ? assessmentWithHistory.changeHistory : []
-    const changeHistoryUpdated = prevChangeHistory !== changeHistory
+  updatePrintButtonIfNeeded(prevAssessmentHistory) {
+    const { assessmentHistory } = this.props
+    const assessmentHistoryUpdated = Boolean(prevAssessmentHistory !== assessmentHistory)
```
wmitchell-cambria: nit: This seems a little verbose to me. `!==` will already be a Boolean, so the cast is unnecessary. (Besides, even if it were not a Boolean, it is only used in the conditional anyways...) Why not just inline it all into the `if` statement?

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
```diff
@@ -38,51 +38,47 @@ class AssessmentChangeLog extends Component {
...
   }
 
   shouldPrintButtonBeEnabled() {
-    const { assessmentWithHistory } = this.props
-    const changeHistory = assessmentWithHistory.changeHistory ? assessmentWithHistory.changeHistory : []
-    const changeHistoryLength = changeHistory.length
-    const enablePrintButton = changeHistoryLength > 0
+    const { assessmentHistory } = this.props
+    const assessmentHistoryLength = assessmentHistory.length
+    const enablePrintButton = assessmentHistoryLength > 0
```
wmitchell-cambria: nit: Just return it here instead of setting it to a variable?

I think some of this code is coming out like this because it is simplifying what used to be more complex. Sometimes it helps to step back and look at the final result with fresh eyes.

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
```diff
@@ -132,10 +125,7 @@ class AssessmentChangeLog extends Component {
 }
 
 AssessmentChangeLog.propTypes = {
-  assessmentWithHistory: PropTypes.shape({
-    changeHistory: PropTypes.arrayOf(changeHistoryPropTypes),
-    assessment: PropTypes.object,
-  }),
+  assessmentHistory: PropTypes.arrayOf(assessmentHistoryPropTypes),
```
wmitchell-cambria: I hadn't realized how much extra logic there was because of this complex prop. It's so clean after these changes!

app/javascript/Application/components/Assessment/assessment.mocks.test.js
```diff
@@ -980,6 +1039,37 @@ export const assessmentCompletedWithReferralNumber = {
...
+  updated_by: {
+    first_name: 'Name 2',
+    last_name: 'LastName 2',
+  },
+  county: {
+    name: 'Alameda',
+  },
+}
+
+export const assessmentDeletedWithReferralNumber = {
```
wmitchell-cambria: I'm starting to think we need a better way of defining these... Maybe we should make a tech debt story? Having thousands of lines of mock data like this can be tedious to manage, as a change in a particular field needs to be applied to each one. It also hides a lot of details from the tests themselves, and those details may or may not be important.

It's possible that something like FactoryBot would be helpful here, where we could construct mock assessments using descriptive code. `assessment.withStatusDeleted().withReferralNumber()`

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.js
```diff
@@ -1,31 +1,37 @@
...
         <AddCansLink clientIdentifier={clientIdentifier} />
       </AuthBoundary>
-    )
+    ) : null
   }
 
   renderAssessments = assessments => {
     const assessmentsToDisplay = 3
+    const { callback } = this.props
+    const loadingState = this.props.loadingState ? this.props.loadingState : ''
```
wmitchell-cambria: nit: Would something bad happen if `this.props.loadingState` were not defined? Why do we default it to an empty string?

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.js
```diff
@@ -84,8 +93,10 @@ class ClientAssessmentHistory extends PureComponent {
 
 ClientAssessmentHistory.propTypes = {
   assessments: PropTypes.arrayOf(PropTypes.object),
+  callback: PropTypes.func.isRequired,
```
wmitchell-cambria: nit: Callback for what? It's required, but there is not much info about when it is called or why, which could help the parent author to decide what function is appropriate to set.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.test.js
```diff
@@ -131,5 +135,16 @@ describe('<ClientAssessmentHistory', () => {
...
     })
+
+    describe('when assessments are loading', () => {
+      it('renders the loading message', () => {
+        // given + when
+        const wrapper = getWrapper([], LoadingState.waiting)
+
+        // then
+        const message = wrapper.find('#no-data').text()
+        expect(message).toBe('Loading assessments...')
```
wmitchell-cambria: Is this related to the Delete story? It seems like handling this loading state roped in a couple hundred lines of code changes that could have been split out. It's harder as a reviewer to focus on a PR that covers multiple topics.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryLoadingBoundary.js
```diff
@@ -4,9 +4,14 @@ import LoadingBoundary from '../../common/LoadingBoundary'
 import { AssessmentService } from '../../Assessment/'
 
 const ClientAssessmentHistoryLoadingBoundary = ({ clientIdentifier, children }) => {
-  const service = () => AssessmentService.search({ client_identifier: clientIdentifier })
+  const service = () =>
```
wmitchell-cambria: Same comment as before. This should not be created each and every render. Only when the identifier changes.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTable.js
```diff
@@ -9,21 +9,28 @@ import ClientAssessmentHistoryTableDate from './ClientAssessmentHistoryTableDate
 import ClientAssessmentHistoryTableUpdatedBy from './ClientAssessmentHistoryTableUpdatedBy'
 import ClientAssessmentHistoryTableEllipsis from './ClientAssessmentHistoryTableEllipsis'
 
-const columnWidths = [190, 250, 140, 170, 260, 35]
+const columnWidths = [0, 190, 250, 140, 170, 260, 35]
```
wmitchell-cambria: I can't say for sure that rendering a 0 width comment just as a backup sorting option is a bad idea... but it _does_ seem odd to me. Are we sure we need to do this? How often are there multiple assessments on the same day, and could the Updated Time be a suitable replacement for those rare cases?

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTableEllipsis.js
```diff
@@ -3,14 +3,31 @@ import PropTypes from 'prop-types'
 import Ellipsis from '../../common/Ellipsis'
 
 const ClientAssessmentHistoryTableEllipsis = ({ original }) => {
-  const { person, id, inheritUrl } = original
-  return id && person ? <Ellipsis clientId={person.identifier} inheritUrl={inheritUrl} assessmentId={id} /> : null
+  const { person, id, metadata, status, inheritUrl, callback } = original
+  const shouldRender = Boolean(person && id && metadata && status && inheritUrl && callback)
```
wmitchell-cambria: Is there a valid situation in which these properties may be falsy?

app/controllers/api/assessments_controller.rb
ggomez-cambriasolutions: yes

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.js
ggomez-cambriasolutions: we already define padding in the sass file, both here and the search assessment history.

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
ggomez-cambriasolutions: agreed

app/javascript/Application/components/Assessment/ChangeLog/AssessmentChangeLog.js
ggomez-cambriasolutions: right? Igor made some improvements to the API, so we didn't have to fetch the assessment and the assessment history with two API calls.

app/javascript/Application/components/Assessment/assessment.mocks.test.js
ggomez-cambriasolutions: Yes, we definitely need to clean up this file. It has gotten way too big.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistory.js
ggomez-cambriasolutions: nothing bad would happen. I was just being over cautious, but it's fine just destructuring the loadingState from the props object.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTable.js
ggomez-cambriasolutions: I had to add the assessment id column, because the table without it, after deleting assessment, the assessment record would move to a different row within the same event date grouping, so I had to sort by event date, and then assessment id. That is the purpose of having the hidden column, because the design does not state to show that column. Using the updated by date wouldn't solve this problem of having the record show up in a different spot after deleting and fetching assessment history again. It's not a backup sorting option, it is always used.

wmitchell-cambria: Ah, interesting, that makes sense.  Would another option be to use Created timestamp as the secondary sort field, which wouldn't change when the assessment is deleted? (By 'backup', I just meant that it doesn't affect the sort unless the assessment dates are the same. It's a secondary field.)

app/javascript/Application/components/Search/SearchAssessmentHistory.js
```diff
@@ -3,22 +3,30 @@ import PropTypes from 'prop-types'
...
+import { AssessmentStatus, sortAssessments } from '../Assessment'
 
 class SearchAssessmentHistory extends Component {
   constructor(props) {
     super(props)
-
     this.state = {
       assessments: [],
       fetchStatus: LoadingState.idle,
+      lastDeletedAssessmentId: 0,
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: The assessment Id is an integer. I made a loading boundary for search assessment history, to help with updating the assessment history when we delete an assessment on the search page. So, we don't need lastDeletedAssessmentId in state anymore.

app/javascript/Application/styles/global/colors.sass
```diff
@@ -31,6 +32,7 @@ $burgundy: #AD2213
 
 // Black
 $transparent-black: rgba(0, 0, 0, .7)
+$transparent-black-light: rgba(0, 0, 0, .2)
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: you know, it's like black, but lighter. :)

app/javascript/Application/components/common/AssessmentRecordStatus.test.js
```diff
@@ -10,9 +10,9 @@ describe('AssessmentRecordStatus', () => {
         status: AssessmentStatus.inProgress,
       })
       expect(expectedResult).toEqual(
-        <div className="status-icon-wrapper">
-          <Icon name="circle-notch" set={'fa'} className={`fa-2x`} />
-          <span className="assessment-in-progress">In Progress</span>
+        <div className={'status-icon-wrapper'}>
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: I didn't see any lint warnings for this component. What are your thoughts on this?

app/javascript/Application/components/common/Ellipsis.js
```diff
@@ -23,31 +26,139 @@ export default class Ellipsis extends React.Component {
...
+
+    return isDeleteAssessmentWarningShown ? (
+      <PageModal
+        isOpen={true}
+        title={'Deleting CANS Warning'}
+        warningDescription={warning}
+        description={'This cannot be undone.'}
+        removeButtonLabel={'Delete CANS'}
+        cancelButtonLabel={'Cancel'}
+        onCancel={() => this.handleWarningCancel()}
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: I redefined the methods that are passed as props within Ellipsis as anonymous functions. Is this what you meant, or is it always better to define the methods as anonymous functions?

app/javascript/Application/components/common/Ellipsis.js
```diff
@@ -23,31 +26,139 @@ export default class Ellipsis extends React.Component {
...
+  async handleWarningDelete() {
+    const { assessmentId } = this.props
+    const { callback } = this.props
+    await AssessmentService.delete(assessmentId)
+    this.closeModal()
+    callback(assessmentId)
+  }
+
+  canDeleteAssessment() {
+    try {
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: another case of being overly cautious. I refactored this piece of logic to use null checks, instead.

app/javascript/Application/components/common/Ellipsis.js
```diff
@@ -23,31 +26,139 @@ export default class Ellipsis extends React.Component {
...
+
+  renderViewChangeLogMenuItem() {
+    return (
+      <div
+        className={'popover-view-change-log'}
+        onClick={e => this.onItemSelect(e)}
+        onKeyPress={e => this.handleKeyPress(e)}
+        role={'menuitem'}
+        aria-label="View Change Log Menu Item"
+        tabIndex={0}
```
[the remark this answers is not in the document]
ggomez-cambriasolutions: Yes, you are correct. It makes more sense to use a button element for each menu item. Done. :)

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTableEllipsis.js
ggomez-cambriasolutions: Good point. If the Ellipsis component is being rendered, then that means there is an assessment  associated with it. So, there really isn't a valid situation that these props would be falsy. Otherwise we wouldn't be rendering an Ellipsis component in the first place.

app/javascript/Application/components/common/Ellipsis.test.js
ggomez-cambriasolutions: you're right, @pokharelmanish . popover is one word. :)

app/javascript/Application/components/common/Ellipsis.js
ggomez-cambriasolutions: great suggestion, @pokharelmanish

app/javascript/Application/components/common/Ellipsis.js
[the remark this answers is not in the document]
wmitchell-cambria: While I think of it, I should give two actual reasons why the null checks are better than the try/catch:

1. Performance can be slow on certain engines. For example, V8 used to skip optimizing functions that had try/catch. Chrome handles this well now, but that is only one of the ~5 engines we need to support.
2. Linear flow. A try/catch represents a possibility that any command will jump you down to the catch. Handling specific errors with null checks lets you read from left to right, top to bottom. That plays into complexity metrics for Code Climate, for example.

app/javascript/Application/components/Assessment/ChangeLog/ChangeLogLoadingBoundary.js
wmitchell-cambria: The loading boundary changes look good!

PR #402, opened 2019-01-03
CANS-879 fix styles and wording

abelayt:
## Tests

## Types of changes

## Checklist:


abelayt:
retest this please

app/javascript/Application/styles/global/cwds-conflict-fixes.sass
```diff
@@ -46,6 +46,7 @@ button
 
 .ReactTable .-pagination .-pageSizeOptions
   margin: 0 1rem !important
+  width: 12rem
```
wmitchell-cambria: 😢 Another reason to remove React Wood Duck.

package.json
```diff
@@ -16,7 +16,7 @@
     "test:a11y": "bundle exec rspec spec/a11y"
   },
   "dependencies": {
-    "@cwds/components": "^0.5.9",
+    "@cwds/components": "^0.6.0",
```
AK56: In '0.6.0', component lib updated casing of Infotips from 'InfoTip' to 'Infotip' and I would expect yarn install to update node_modules in the app accordingly, but it doesn't seem it did, which failed the web-pack build with the following error
cans-879module_not_found

Wondering if anybody else encountered it or am I missing something ? @abelayt  @wmitchell-cambria @denys-davydov

wmitchell-cambria: I feel like I noticed some similar behavior, but I'm not sure. I was deep into a package dependency nightmare yesterday for other reasons. I could be confusing the two.

This could be a bug in yarn, or a compatibility issue with Mac's case-insensitive file system. When that happens, I usually just blow away my node_modules and install fresh (which is similar to what Jenkins ends up doing).

abelayt: I've informed @eckdanny-osi to take a look at this error.

denys-davydov: My 2 cents: @AK56 , if this build is green then you probably have local issues. Try to delete _node_modules_ folder manually as Will described.

AK56: Hmm Mac's file system case insensitivity shouldn't have let to this error then, since it would take 'InfoTip' and 'Infotip' as the same, or I wonder if the case insensitivity didn't allow yarn to update the naming of the module on disk, while the build is expecting the new name. 
Anyways, a fresh build of node_modules fixes it thanks @wmitchell-cambria

wmitchell-cambria: > I wonder if the case insensitivity didn't allow yarn to update the naming of the module on disk, while the build is expecting the new name.
That would be my guess as well.

PR #411, opened 2019-01-07
CANS-915 regression test script for Case worker login and Assessment in process.

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-915

## Description
Creating test script to automate an acceptance script in CANS so that I can test a	Case worker login and Assessment in-process.

**Testing**
In order to test this script please use the user id and password shared earlier and update the file - **run-regression-test.sh** and update "test:regression" in package.json to point to index page inside regression folder.

## Tests

## Types of changes

## Checklist:


spec/acceptance_helpers/resource_helper.rb
```diff
@@ -345,6 +353,34 @@ def fetch_traumatic_domain
...
+    find('#age-0-5-button').click # doing it twice since sometimes it fails on the first attempt
+    find('#has-caregiver-yes').click
+    find('#can-release-control').click
+    expect(page).to have_content('By selecting NO, Items 7, 48, and EC 41 (Substance Use Disorder Items) from this CANS assessment will be redacted when printed.', wait: 60)
+    fill_assessment_form_domains_for_age_0_5
+  end
+
+  def fill_assessment_form_domains_for_age_0_5
+    fetch_challenges_domain
+    # this is a temporary comment for other dev working on the same scripts.
```
nasar-cambria: @zliuwd I commented this out with an intention to use this for filling up domains. Also, we make use of other methods in this file or modify them as necessary.

nasar-cambria:
re-test this please

spec/acceptance_helpers/resource_helper.rb
zliuwd: There are two things:
1. For your story, it just need you fill out 1 or 2 items of the first domain, you do not need fill out more than that
2. Among all of these regression test stories, only two of them need we fill-out all the domains, so we can use the algorithm in my PR to avoid this "hard code ID block". 
Oppositely, if we just need select couple elements, sure we could use couple ids..

nasar-cambria:
@zliuwd Thats is correct if you see my pr it is not ready to be reviewed (WIP). I created this just for reference so we can collaborate on it and not do repeated work. 

To your second question. There is no hard code ID block here I am using clinet's name as an identifier to navigate back to client page.

spec/acceptance_helpers/resource_helper.rb
```diff
@@ -12,6 +12,14 @@ def client_identifier
     '0PcpFQu0QM'
   end
 
+  def post_new_assessment_regression
+    verify_assessment_age_0_5
+  end
+
+  def caseworker_client_identifier
```
LeonidMarushevskyi: Let's make an actionable name for this step and pass client name as a parameter so that this step will be reusable like that click_caseworker_client('Case, Child 01 Test, Suff')

spec/regression/index_page_spec.rb
```diff
@@ -8,6 +8,10 @@
     login
     expect(page).to have_content('CANS')
     expect(page).to have_content('Client List')
+    caseworker_client_identifier
```
LeonidMarushevskyi: You should rename scenario file as needed for your scenario.

LeonidMarushevskyi: Scenario name needs to be changed to what you whan to achieve. 
Currently it is scenario 'CANS-Worker lands on client list and can logout'

spec/acceptance_helpers/resource_helper.rb
```diff
@@ -345,6 +353,24 @@ def fetch_traumatic_domain
     traumatic_domain.each { |element| find(element).click }
   end
 
+  def verify_assessment_age_0_5
```
LeonidMarushevskyi: We should split this method in smaller one responsibility methods and use them in the scenario. Keep small reusable steps in helper and move all specifics to actual scenario. Now it looks look very specific case which will be hard to reuse.

spec/acceptance_helpers/resource_helper.rb
```diff
@@ -12,6 +12,14 @@ def client_identifier
     '0PcpFQu0QM'
   end
 
+  def post_new_assessment_regression
```
LeonidMarushevskyi: Why we need this function which just wraps another function?

nasar-cambria: removed. I am now calling the method directly.

spec/acceptance_helpers/resource_helper.rb
nasar-cambria: done, I made this to just test radios in the header which can be reusable.

spec/acceptance_helpers/resource_helper.rb
```diff
@@ -345,6 +353,24 @@ def fetch_traumatic_domain
...
+    find('#can-release-control').click
+    substance_use_disorder_text = [
+      'By selecting NO, Items 7, 48, and EC 41',
+      '(Substance Use Disorder Items) from this CANS assessment',
+      'will be redacted when printed.'
+    ].join(' ')
+    expect(page).to have_content(substance_use_disorder_text, wait: 45)
+    fetch_challenges_domain
+    click_button 'Save'
+    caseworker_client_identifier # doing it twice since sometimes it fails on the first attempt
```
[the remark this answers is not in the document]
nasar-cambria: We need to remove this and replace it with what @zliuwd  has in his code and use that to visit client page. So, except this one I pushed all other changes.

cc: @LeonidMarushevskyi

PR #414, opened 2019-01-08
CANS-913-SupervisorLoginRegressionTest

AK56:
## Description
Creating test script to automate an acceptance script in CANS so that I can test a Supervisor login, checkout his staff list and clients of one of his staff.

Testing
In order to test this script please use the supervisor user id and password shared earlier.

## Tests

## Types of changes

## Checklist:


spec/acceptance_helpers/resource_helper.rb
```diff
@@ -16,6 +16,10 @@ def caseworker_client_identifier
     find(:link, 'Case, Child 01 Test, Suff', wait: 15).click
   end
 
+  def visit_staff_member_dashboard
+    find(:link, 'Regression, QA02', wait: 15).click
```
LeonidMarushevskyi: Let's pass 'Regression, QA02' as the parameter to this function like 
def visit_staff_member_dashboard(staff_member_name)
 find(:link, staff_member_name, wait: 15).click
end

In such way, this function will be reusable if someone will want to click other subordinate name. Let's keep scenario-specific data like staff person name in scenario.

spec/regression/supervisor_login_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+feature 'Index Page' do
+  scenario 'Supervisor lands on staff list, select a staff member, visits one of staff member clients and logs out' do
+    login supervisor_json
+    expect(page).to have_content('CANS')
+    expect(page).to have_content('Assigned Staff', wait: 10)
+    visit_staff_member_dashboard
+    expect(page).to have_content('Client List', wait: 10)
+    visit_client_profile('Case, Child 01 Test, Suff')
+    expect(page).to have_content('Client Information')
+    expect(page).to have_content('Assessment History')
```
wmitchell-cambria: We should probably verify something about the client info appearing, like the name. Otherwise, we could run into an issue where we navigate to an empty Client Profile page, or one for the wrong Client.

spec/acceptance_helpers/resource_helper.rb
AK56: Good call! I have updated that piece, and also alongside did some refactoring moving the CLIENT_NAME and STAFF_NAME constants in our spec/acceptance_helper file, so they can be shared across our spec files.

spec/acceptance_helper.rb
```diff
@@ -10,6 +10,9 @@
 require 'acceptance_helpers/login_helper'
 require 'acceptance_helpers/prod_login_helper'
 
+CLIENT_NAME = 'Case, Child 01 Test, Suff'
```
LeonidMarushevskyi: Let's keep scenario specific data like this constants in actual scenario instead of helper.

spec/regression/supervisor_login_spec.rb
LeonidMarushevskyi: Let's make sure that all Expected Outcomes from Testing Matrix are covered with expect statements.

AK56: Agree

spec/acceptance_helper.rb
AK56: Keeping the constants in a helper that we are including across our spec files makes our pieces reusable. Please read my comment above

AK56:
retest this please

PR #428, opened 2019-01-12
CANS-897: Staging regression test Jenkins file added

LeonidMarushevskyi:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-897

## Description

Additional Jenkins file created for Regression Test in Staging. Currently, Production Jenkins does not support shared Jenkins library so was not able to use the main Jenkins file. Later Jenkinsfile_staging_regression.groovy can be deleted and the main Jenkins file can be used.

## Tests

## Types of changes

## Checklist:


LeonidMarushevskyi:
Jenkins retest this please

spec/acceptance_helpers/prod_login_helper.rb
```diff
@@ -44,8 +44,8 @@ def supervisor_json
   end
 
   def enter_verification_code(login_config)
-    return unless page.has_xpath?('//label[text()="Verification Code"]')
-    fill_in 'Verification Code', with: login_config[:verification_code]
+    return unless page.has_content?('Verification')
```
LeonidMarushevskyi: Staging and Production environments have an older version of the login page. Reworked method to be backward compatible and work in Int and on Staging.

LeonidMarushevskyi:
Jenkins retest this please

Jenkinsfile_staging_regression.groovy
```diff
@@ -0,0 +1,53 @@
+//TODO: remove this pipeline file and use main Jenkins file when Prod Jenkins will support shared Jenkins library
```
wmitchell-cambria: Is there a timeline for when Prod Jenkins will support the shared library? Or perhaps a Jira story to track it?

docker-compose.ci.yml
```diff
@@ -143,6 +143,8 @@ services:
       CANS_WEB_BASE_URL: 'cans-web:3000/cans'
     tty: true
     stdin_open: true
+    extra_hosts:
+      - "staging.cwds.ca.gov: 10.211.81.217"
```
wmitchell-cambria: I am concerned about hardcoding an IP address here. IP addresses can change, especially ones in private address spaces like `10.0.0.0/8`. Is there a way to rely on DNS instead, so that if `staging.cwds.ca.gov` changes to a new IP address, the regression tests can verify that it is still working as expected?

Additionally, while not strictly sensitive information, this piece of configuration is specific to running the app in our ecosystem. Most similar variables in this file are included as environment variables (eg, db hosts, schema names, etc). Should this be included as an environment variable as well?

spec/acceptance_helpers/prod_login_helper.rb
wmitchell-cambria: I think this makes sense given our current structure. As a side note, I have been reading more about [Twelve-Factor Apps](https://12factor.net/) lately. One aspect of them is that all code is stored in a single repository, and dependencies are explicitly declared. That would be one way to solve this situation of version drift between applications. (Essentially, right now we need to be able to consume two different 'APIs' of the login page.)

Anyways, that's more just for reading than any actual suggestion. It just struck me as an interesting tangent.

Jenkinsfile_staging_regression.groovy
LeonidMarushevskyi: Task created for team H https://osi-cwds.atlassian.net/browse/CANS-941. Don't have a timeline for this yet. But we are not blocked and change will be fairly simple on our side.

docker-compose.ci.yml
LeonidMarushevskyi: After consultation with DevOps IP removed.

PR #430, opened 2019-01-14
CANS-914 Regression test - Non Case worker

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-914

## Description
Regression test script in CANS to test a Non-case worker login.

**Testing**
In order to test this script please use the user id and password shared earlier and update the file - **run-regression-test.sh** and update "test:regression" in package.json to point to ```run-regression-test``` file

## Tests

## Types of changes

## Checklist:


spec/page_objects/client_search.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'acceptance_helper'
+
+class ClientSearch < SitePrism::Page
+  element :close_top_alert, 'i.close-icon'
+  element :search_clients, 'input#downshift-0-input'
+  element :search_clients_link, 'div.full-name', text: 'Case, Child 01 Test'
```
LeonidMarushevskyi: Probably name found_client_link will speak better to the element of the page. "Case, Child 01 Test" - is a client name for a particular scenario. We need to keep elements generic so that we can use them to find and open any client.

spec/page_objects/client_search.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'acceptance_helper'
+
+class ClientSearch < SitePrism::Page
+  element :close_top_alert, 'i.close-icon'
```
LeonidMarushevskyi: This probably should be named top_alert_close_icon

spec/page_objects/client_search.rb
```diff
@@ -0,0 +1,9 @@
...
+  element :search_clients, 'input#downshift-0-input'
```
LeonidMarushevskyi: This is search_input

spec/page_objects/client_search.rb
LeonidMarushevskyi: You need to define a function which will use this element to find the proper client and click it.

spec/regression/non_case_worker_scenarios_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  scenario 'Non Case worker login, search client, select it and logs out' do
+    login non_caseworker_json
+    expect(page).to have_content('To Start a CANS Assessment, Search and Select the Child')
+    @form.close_top_alert.click
+    expect(page).not_to have_content('To Start a CANS Assessment, Search and Select the Child')
+    @form.search_clients.click
+    @form.search_clients.set CLIENT_NAME
+    expect(@form.search_clients_link).to have_content('Case, Child 01 Test')
+    @form.search_clients_link.click
+    expect(page).to have_content 'Assessment History'
```
zliuwd: According to the spreadsheet #14. Select a child from the list of search results, the outcome as below:
**Routed to the matching Client Information/Assessment History.**
Here we only verified the title 'Assessment History' is shown. We also should verify client information is matching.
For example: try to verify the first and last name are matching the client's name which we searched.

spec/page_objects/client_search.rb
nasar-cambria: ```close_top_alert``` - Since it's a variable I am defining it for it's action. The action rule would be easy to understand for readability. Please let me know if you have other ideas.

PR #434, opened 2019-01-15
CANS-932. Improve performance for Read Staff Latest Assessments operation

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-932

## Description
The main performance improvement work for this task is made on the backend within the following PR https://github.com/ca-cwds/cans-api/pull/201 
With this PR the frontend processing of the assessments (like filtering by status and sorting by timestamps) is removed, the data is prepared on the backend fully.

## Tests

## Types of changes

## Checklist:


config/routes.rb
```diff
@@ -37,7 +37,7 @@
     resources :staff, only: [:show] do
       collection do
         get 'subordinates', to: 'staff#subordinates_index'
-        get 'assessments', to: 'staff#assessments'
+        get 'assessments/latest', to: 'staff#assessments'
```
wmitchell-cambria: I think... unless we expose another way to fetch assessments for a staffperson, we could leave this Rails endpoint at `'assessments'`, and only change the `StaffRepository`. That would avoid the 'conversion' from `assessments/latest` (Rails endpoint and route) to `assessments` (Rails controller method) and back to `assessments/latest` (CANS API endpoint).

Alternatively, we could name it `assessments/latest` or `assessments_latest` all the way through. (I think there is a better naming scheme for this, but I am not clear on it.)

app/javascript/Application/components/Search/AssessmentHistory/SearchAssessmentHistory.js
```diff
@@ -2,62 +2,34 @@ import React from 'react'
...
-
-  renderAssessments() {
-    const { loadingState, navFrom, inheritUrl } = this.props
-    const assessments = this.processAssessments()
-    const noDataText =
-      loadingState === LoadingState.ready ? 'No assessments currently exist for the clients.' : 'Loading assessments...'
-
-    return assessments.length === 0 ? (
-      <div id="no-data">{noDataText}</div>
+const SearchAssessmentHistory = props => (
```
wmitchell-cambria: I suspect now that this is simplified a bit, that we can share much more code between here and ClientAssessmentHistory. I'll look into it a bit and create a tech debt story if it looks worthwhile.

denys-davydov: The sorting function that was previously used here is still needed for at least one other component, so it should stay for now.

wmitchell-cambria:
Removing the blocker, per comments on the backend PR. Note that Jenkins is failing due to a code coverage threshold. We probably need to add back a few tests.

app/javascript/Application/components/Search/AssessmentHistory/SearchAssessmentHistory.js
denys-davydov: @wmitchell-cambria  And I will refactor the function so the tech debt story will not be needed.

wmitchell-cambria:
Refactoring changes look OK.

PR #441, opened 2019-01-18
CANS-920 Delete and Change Log test scenarios.

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-920

## Description
Scenario 63: User is routed to the CANS Change Log and is able to view any changes made to this CANS assessment.  Change Log displays the time/date the assessment was last updated, who made the change, and the status of the change. 

**Please Note:** As discussed with PO for the above test case scenario we are only validating Client name and Assessment Date. 

**Testing**
In order to test this script please use the user id and password shared earlier and update the file - **run-regression-test.sh** and update "test:regression" in package.json to point to index page inside regression folder.

## Tests

## Types of changes

## Checklist:


nasar-cambria:
Jenkins re-test this please

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -87,6 +91,25 @@
     save_and_check_the_success_message
   end
 
+  def delete_assessment_test(current_date)
+    @client_profile.ellipsis_icon.click
+    click_button 'Delete CANS'
```
wmitchell-cambria: This operates on the page without using the page object.

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -87,6 +91,25 @@
...
+    expect(@form.global.delete_warning_modal['style']).to eq('display: block;')
+    @form.global.cancel_button_of_warning.click
+    expect(page).to have_content('Assessment History')
```
wmitchell-cambria: We should start aiming to replace these types of expectations with URL matchers, perhaps.

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -87,6 +91,25 @@
...
+    @client_profile.ellipsis_icon.click
+    click_button 'Delete CANS'
```
wmitchell-cambria: Here, also, operating on the page without using the page object.

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -87,6 +91,25 @@
...
+    @form.global.agree_button_of_warning.click
+    expect(page).to have_content('Deleted on ' + current_date)
+  end
+
+  def view_cans_change_log_test(current_date)
+    @client_profile.ellipsis_icon.click
+    click_button 'View CANS Change Log'
```
wmitchell-cambria: Same here, operating on the page without using the page object.

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -69,12 +69,16 @@
...
     click_button 'Save'
+    current_date = Time.now.strftime('%m/%d/%Y')
     @form.global.wait_until_global_save_success_message_box_visible
     @form.breadcrumbs.route_from_breadcrumbs(CLIENT_NAME_2)
     expect(@client_profile).to have_in_progress_record
-    current_date = Time.now.strftime('%m/%d/%Y')
     find(:link, current_date + ' CANS', match: :first).click
     expect(page).to have_content 'CANS Communimetric Assessment Form'
+    @form.breadcrumbs.route_from_breadcrumbs(CLIENT_NAME_2)
+    view_cans_change_log_test(current_date)
```
LeonidMarushevskyi: Nasar, let's create a separate scenario for your task instead of modify existing scenario.

nasar-cambria: I feel that it fits in this scenario. We need to duplicate most of the scenarios from this scenario in order to test an assessment change log and delete functionality. I would suggest we keep it as is. I don't mind having a separate scenario but it would repeat the whole process from Start to the End (login, create and save assessment). This is going to increase the overall time of running entire test case which keeps growing every-time we create new assessment and increases load on the data base as well.

cc: @wmitchell-cambria @AK56

wmitchell-cambria: Generally, these _are_ very slow tests. If they can be separated without sharing a lot of setup, then that tends to be good (because they can theoretically run in parallel). If they have to share setup, keeping them in a single scenario is just so much faster. In this case, I could definitely see it going either way. (I'm not so sure how long the setup takes at the moment.)

In any case, if we keep this changelog and delete behavior in this scenario, at the very least the scenario description needs to change to be more general. I would not expect "login, create, logout" to delete the assessment.

spec/regression/case_worker_scenarios_spec.rb
AK56: Are you referring to use something like client_profile_path in this case ? Which I'm totally down for, but I thing it requires client id, which we decided on not to use in our regression suite.

spec/regression/case_worker_scenarios_spec.rb
LeonidMarushevskyi: This scenario already complex and unreadable. You don't need to repeat all steps. Your scenario can be as simple as this: 

CURRENT_DATE = Time.now.strftime('%m/%d/%Y')
scenario 'Assessment change log and delete' do
    login    
    create_new_assessment(CLIENT_NAME) // need to parameterize this existing  function
    click_0_to_5_button
    save_and_check_the_success_message
    @form.breadcrumbs.route_from_breadcrumbs(CLIENT_NAME) // can be moved to separate function in assessment_form.rb
    view_cans_change_log_test(CURRENT_DATE)
    @form.breadcrumbs.route_from_breadcrumbs(CLIENT_NAME)
    delete_assessment_test(CURRENT_DATE)
  end

AK56: I feel like we don't have hard rules set to define what each scenario should be. This is something that got derived from the excel sheet which basically told us to make sure we are covering/validating those tests. Although we have done our best to have a separate scenario to test a functionality that is very complex and shouldn't overlap with other part of the app. 
Given the modularity of 'delete' assessment test cases, and that it will expedite our testing I feel we can integrate it into an existing scenario.

nasar-cambria:
Thanks all for your review. Please feel free to re visit and release the block when you get a chance.

CC: @LeonidMarushevskyi @wmitchell-cambria

spec/regression/case_worker_scenarios_spec.rb
LeonidMarushevskyi: Even if you all insist on using one big scenario, it should be refactored. Currently, it violates Single Level of Abstraction principle (http://www.principles-wiki.net/principles:single_level_of_abstraction) which makes it harder to comprehend. In this scenario we have high-level functions which aggregate steps together with low level functions like find(:link, current_date + ' CANS', match: :first).click.

We want to keep scenarios compact and user readable so that they can be reviewed by business people. Back to the example above in my comment where inside scenario all functions are on the same level of abstraction and lower level functions extracted to separate small functions. It makes code readable, functions small.

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -69,12 +69,16 @@
     verify_radio_buttons_on_assessment_header
     validate_domain_radio_and_chevron
     click_button 'Save'
+    current_date = Time.now.strftime('%m/%d/%Y')
```
LeonidMarushevskyi: Extract this into constant. It is used in one more place in this class.

spec/regression/case_worker_scenarios_spec.rb
LeonidMarushevskyi: Site prism has matchers to verify that particular page is displayed https://github.com/natritmeyer/site_prism#verifying-that-a-particular-page-is-displayed

LeonidMarushevskyi:
@nasar-cambria what about refactoring of scenario (my comment above)? If you touching the existing scenario, please spend some time and refactor the scenario to use prism and make it more readable.

spec/regression/case_worker_scenarios_spec.rb
nasar-cambria: We need a different expect statement here which can validate if the page modal is closed. Changes updated in the clean-up commit.

For this scenario I am closing the page modal and just trying to validate that I have returned to the previous view which is Assessment History table. So, the expect statement is to make sure the modal is closed and I am back to the previous view. Using prism ```Testing for Page display``` would test which page we are at but not it's view. You can be on the same page but your view could be different. Ex: The page modal could still be viewable or some other pop up might be on the page.

spec/regression/case_worker_scenarios_spec.rb
nasar-cambria: What you are suggesting here is about the scenario which was previously created and refactored by @zliuwd. If we need any further refactoring let us discuss once and I will ensure that your concerns are addressed. Thanks!

spec/regression/case_worker_scenarios_spec.rb
nasar-cambria: done

spec/regression/case_worker_scenarios_spec.rb
zliuwd: What @LeonidMarushevskyi  mentioned here has noting about my previous scenario. What he pointed out is: in scenario 'Case worker creates and deletes assessment', "wrapped up functions(high level)" and "original test lib functions(low level)" are mixed using. For example, if we look at Gurpartap's  work (as below), we will found he only use wrapped up (high level) functions in it.
```
scenario 'Case worker login, tests caregiver domain with new assessment and logs out' do
    login
    create_new_assessment
    click_0_to_5_button
    expand_all_domains
    caregiver_domain_name_and_item_rating_test
    has_caregiver_domain_option_yes_to_no_test
    save_and_check_the_success_message
  end
```

nasar-cambria:
Closing this PR as we have everything included in the one below.
https://github.com/ca-cwds/cans/pull/445

PR #443, opened 2019-01-18
CANS-725 - Item Rating refactor

wmitchell-cambria:
## JIRA
Followup refactors for [CANS-725 - Fix eslint warnings](https://osi-cwds.atlassian.net/browse/CANS-725)

## Description

This has three main parts:
1. Refactor ItemRegularRating and ItemBooleanRating to share a _lot_ more of their code through a base ItemRating component.
2. Move Item-related files to their own directory within `/Assessment`, because there are a lot of them now.
3. Factor various definitions of `const naRating = 8` into a shared RatingHelper file to avoid some hard-coded values.

## Tests

## Types of changes

## Checklist:


wmitchell-cambria:
Dear Reviewers, this may be easier to look at commit-by-commit, because of the file renaming.

app/javascript/Application/components/Assessment/Item/ItemRegularRating.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ItemRating from './ItemRating'
+
+const options = ['0', '1', '2', '3']
```
zliuwd: I think these options arrays is also good to move to RatingHelper, then they also could reuse in some test.

app/javascript/Application/components/Assessment/Item/ItemBooleanRating.js
```diff
@@ -0,0 +1,30 @@
...
+  itemCode: PropTypes.string.isRequired,
+  onRatingUpdate: PropTypes.func.isRequired,
+  rating: PropTypes.number.isRequired,
+}
+
+ItemBooleanRating.defaultProps = {
+  disabled: false,
+}
+
+export default ItemBooleanRating
```
AK56: The `ItemBooleanRating` and `ItemRegularRating` components have lot of duplication besides them setting `type` and `shortType` props differently on ItemRating. We can replace these two components by using `ItemRating` directly in our Item.js and tweak `type` and `shortType` props based on rating_type

app/javascript/Application/components/Assessment/Item/ItemRegularRating.js
```diff
@@ -0,0 +1,30 @@
...
+  itemCode: PropTypes.string.isRequired,
+  onRatingUpdate: PropTypes.func.isRequired,
+  rating: PropTypes.number.isRequired,
+}
+
+ItemRegularRating.defaultProps = {
+  disabled: false,
+}
+
+export default ItemRegularRating
```
AK56: See my comment on ItemBooleanRating.js

app/javascript/Application/components/Assessment/Item/ItemRating.js
```diff
@@ -16,34 +17,29 @@ const PrimRadio = withStyles({
...
+          }}
+        />
+      }
+      label={label}
+      labelPlacement={'start'}
+    />
+  ))
   return (
-    <div className={'item-bool-rating'} style={{ display: 'flex' }}>
+    <div className={`item-${props.shortType}-rating`} style={{ display: 'flex' }}>
```
zliuwd: I think using shortType will keep the className same as old className, then this will not break test right?

app/javascript/Application/components/Assessment/Item/ItemBooleanRating.js
wmitchell-cambria: Possibly, yeah. This gets most of the complicated parts de-duplicated (children, classes, mapping), without actually changing the unit tests at all. That seemed safer to me. Adding knowledge of 'options' to the Item would increase its responsibility, which I'm not crazy about. An alternative to trim this down further would be to pass the props through with a spread operator, since we are just adding new props. We wouldn't lose any proptype validation, because the ItemRating still validates. I'll add that in.

wmitchell-cambria:
@AK56 - Try this on for size. I refactored to remove some of the shared boilerplate. I agree that they could probably be further simplified, but I am very hesitant to touch `Item.js` as part of this PR. It is a very large file, and I don't want to risk it if we can keep the refactor contained.

PR #444, opened 2019-01-19
CANS-938 Refactor Ellipsis Popover Components

ggomez-cambriasolutions:
https://osi-cwds.atlassian.net/browse/CANS-938

## Description
This PR refactors all the logic that was embedded into the Ellipsis component, and places the logic in the appropriate components.

Previously, the Ellipsis component was doing more than just "Ellipsis" related work, so this will make each component smaller and more specialized.

This is all related to deleting assessments from the Ellipsis component(s).

This is a tech-debt story that I agreed to take on at the request of @wmitchell-cambria.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTableEllipsis.js
```diff
@@ -1,20 +1,57 @@
...
+      metadata,
+      status,
+      inheritUrl,
+      updateAssessmentHistoryCallback,
+    } = this.props.original
+    const { isDeleteAssessmentWarningShown } = this.state
+
+    return (
+      <React.Fragment>
+        {this.renderAssessmentDeleteModal(isDeleteAssessmentWarningShown)}
```
denys-davydov: Do we really want to render an Assessment Delete Modal even when it is not needed to be rendered? Will it work if we don't render it at all in such cases like this? 
`{isDeleteAssessmentWarningShown && this.renderAssessmentDeleteModal()}`
With that, we do not go deeper in calculations and simply stop here with our modal since it is not needed.

app/javascript/Application/components/common/AssessmentDeleteModal.js
```diff
@@ -0,0 +1,49 @@
...
+    const title = 'Deleting CANS Warning'
+    const warning = <div>You are attempting to delete this CANS.</div>
+    const description = 'This cannot be undone.'
+    const cancelLabel = 'Cancel'
+    const removeLabel = 'Delete CANS'
+
+    return isShown ? (
+      <PageModal
+        isOpen={isShown}
+        title={title}
```
denys-davydov: Maybe it is a matter of personal preferences, but I would prefer to have `title, warning, description, cancelLabel, removeLabel` values to be used inline as string literals instead of by variables that are initialized 5 lines above. It will improve the readability of the code.

app/javascript/Application/components/common/AssessmentHistoryPopover.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Redirect } from 'react-router-dom'
+import { Popover, PopoverBody } from '@cwds/reactstrap'
+import { AssessmentStatus } from '../Assessment/'
+
+class AssessmentHistoryPopover extends React.Component {
+  state = {
+    redirection: {
```
denys-davydov: I am not sure there is a reason to have a `redirection` "wrapper" here, I think `state = { shouldRedirect: false }` will work for us even better

app/javascript/Application/components/common/AssessmentRecordInfo.js
```diff
@@ -94,9 +97,34 @@ class AssessmentRecordInfo extends Component {
...
+  }
+
   render() {
     const { assessment, header } = this.props
-    return this.renderAssessmentInfo(assessment, header)
+    const { isDeleteAssessmentWarningShown } = this.state
+
+    return (
+      <React.Fragment>
+        {this.renderAssessmentDeleteModal(isDeleteAssessmentWarningShown)}
```
denys-davydov: The same situation as above is here with the modal

app/javascript/Application/components/common/Ellipsis.js
```diff
@@ -1,140 +1,44 @@
...
 import PropTypes from 'prop-types'
-import { Popover, PopoverBody } from '@cwds/reactstrap'
 import Icon from '@cwds/icons'
-import { Redirect } from 'react-router-dom'
 import { Button } from '@cwds/components'
-import PageModal from '../common/PageModal'
-import { AssessmentService, AssessmentStatus } from '../Assessment/'
+import AssessmentHistoryPopover from './AssessmentHistoryPopover'
 
 class Ellipsis extends PureComponent {
```
denys-davydov: Do we need this Ellipsis abstraction and is it a real abstraction here? To me, it looks too concrete to be a common component, that is supposed to be highly reused by other components. The real factored out abstraction you've created is `ClientAssessmentHistoryTableEllipsis.js` component, that has the functionality to be reused by `ClientAssessmentHistoryTable.js` and `AssessmentRecordInfo.js` components. To make the components design cleaner, you need:
1. Delete `Ellipsis.js` and move its code into `ClientAssessmentHistoryTableEllipsis.js`
2. Rename `ClientAssessmentHistoryTableEllipsis.js` to something more generic (and probably move to a different folder because it will be used by two other components)
3. Use the renamed component within `ClientAssessmentHistoryTable.js` and `AssessmentRecordInfo.js` components for them to implement Assessment Delete Modal functionality by themselves.

The main reason to do it is not to manage Assessment Delete Modal state in two different components by copy-pasting and to factor it out into one single component.

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTableEllipsis.js
denys-davydov: The comment may become not applicable if you start with this comment first https://github.com/ca-cwds/cans/pull/444#discussion_r249936791

app/javascript/Application/components/common/AssessmentDeleteModal.js
wmitchell-cambria: +1. While I do agree that it's personal preferences... it also adds 6 extra lines to a 10 line function. Concise is good.

app/javascript/Application/components/common/AssessmentHistoryPopover.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Redirect } from 'react-router-dom'
+import { Popover, PopoverBody } from '@cwds/reactstrap'
+import { AssessmentStatus } from '../Assessment/'
+
+class AssessmentHistoryPopover extends React.Component {
```
wmitchell-cambria: This component still feels a little like a bag of functions. It would be great if we could go one step further and separate out the two different buttons into their own encapsulated components. Note how each button has its own render method and its own describe block in tests. Those are good signs that they are really separate.

That being said, it is less valuable if you have to continue to thread the assessment-specific props through. More on that in a future comment, I think.

app/javascript/Application/components/common/AssessmentRecordInfo.js
```diff
@@ -34,7 +37,8 @@ class AssessmentRecordInfo extends Component {
             assessmentMetaData={metadata}
             assessmentStatus={status}
             updateAssessmentHistoryCallback={updateAssessmentHistoryCallback}
-            inheritUrl={this.props.inheritUrl}
+            inheritUrl={inheritUrl}
+            toggleModal={this.toggleModal}
```
wmitchell-cambria: With this refactor, AssessmentRecordInfo now has more responsibility (tracking modal state), and the Ellipsis has to pass through toggleModal and assessment specific information. I'm going to keep reviewing, because I don't have a good solution for this yet... but it definitely seems like a drawback.

app/javascript/Application/components/common/AssessmentRecordInfo.js
wmitchell-cambria: If it's not too complex, my personal preference here is to just render the component directly. Do we need a `renderAssessmentDeleteModal` helper that just builds a props object?

app/javascript/Application/components/common/Ellipsis.js
wmitchell-cambria: I primarily agree.

> To me, it looks too concrete to be a common component, that is supposed to be highly reused by other components.

+1. Look at the props for the component and the children that it is rendering. It is still tightly coupled to the Assessment domain.

In order to achieve the goal of a common 'EllipsisMenu' component, we would have to do a bunch more refactoring. Specifically, some form of dependency injection, where the Assessment-specific children are passed in from above. If done properly, I think this approach could also allow the actions to be entirely self-contained (including their modal state), so that other components like AssessmentRecordInfo wouldn't have to track it themselves. For it to be a common `Ellipsis`, you should not be able to tell by looking at the file whether it would contain Assessment actions, Client actions, or any other set of fun buttons to press.

That sounds much harder, and I think this is still (close to) a good improvement as is. So, I'd put that on the back burner and continue as Denys suggests.

> 1. Delete Ellipsis.js and move its code into ClientAssessmentHistoryTableEllipsis.js

I agree with moving it to the Assessments folder somewhere. I'm not convinced on combining the files. I suppose with the actions factored out into the popover component, combining the CAHTEllipsis and Ellipsis may not be too bad.

> 2. Rename ClientAssessmentHistoryTableEllipsis.js...

Perhaps `AssessmentActionsMenu.js` or `AssessmentActionsEllipsis.js`

> The main reason to do it is not to manage Assessment Delete Modal state in two different components by copy-pasting and to factor it out into one single component.

💯 As mentioned above, the drawback of this refactor is that these components now have an additional responsibility. We should strive to avoid that and encapsulate the modal behavior as much as possible.

app/javascript/Application/components/Assessment/AssessmentDeleteModal.js
```diff
@@ -0,0 +1,42 @@
...
+
+class AssessmentDeleteModal extends React.Component {
+  handleWarningCancel = () => {
+    this.props.toggleModal()
+  }
+
+  handleWarningDelete = async () => {
+    const { assessmentId, toggleModal, updateAssessmentHistoryCallback } = this.props
+
+    await AssessmentService.delete(assessmentId)
```
wmitchell-cambria: This may have existed before, but what happens if you click it multiple times really fast?

app/javascript/Application/components/Client/AssessmentHistory/ClientAssessmentHistoryTable.js
```diff
@@ -19,120 +19,134 @@ const COLUMN_WIDTHS = {
...
-    width: COLUMN_WIDTHS.UPDATED_BY,
-    className: 'text-center',
-    headerClassName: 'text-center',
-    accessor: assessment => {
-      const { updated_by: updatedBy, created_by: createdBy } = assessment
-      const user = updatedBy || createdBy
-      return `${user.first_name} ${user.last_name}`
+    {
+      Header: '',
+      Cell: row => {
```
wmitchell-cambria: I think I just generally dislike column configs. They get so long. My instinct would be to pull this out into its own component instead of defining the SFC here...but that _is_ a lot of overhead. I think I just generally dislike column configs.

app/javascript/Application/components/Assessment/AssessmentDeleteModal.js
ggomez-cambriasolutions: the API returns a 404 error :(

PR #448, opened 2019-01-23
FIT-423 changed checkoutStage to call sh with sudo for removing workspace

osi-richbach:
## JIRA

https://osi-cwds.atlassian.net/browse/FIT-423

## Description
This is a follow up to creating a single release pipeline on the jenkins management server from the main Jenkinsfile.  
Running acceptance and regression tests happens with docker-compose and since docker is run as root, the output into the workspace ( regression-report/* ) gets created as **root:root**.  The next time the build happens, **deleteDir()** fails because that runs as "ansible" which cannot delete the regression-report dir since it is owned by **root:root**.  I changed the call to **deleteDir()** to **sh 'sudo rm -rf *'**
Rather than trying to change the docker compose to run as ansible and deal with any other problems that **might** arise out of that, I went with this for now.  It is really just cleaning up the previous build anyway.


osi-richbach:
retest this please

PR #453, opened 2019-01-24
CANS-944-RefactorRegressionSuiteForPOMCompliance

AK56:
https://osi-cwds.atlassian.net/browse/CANS-944

Refactoring the left outs in our regression test suite, definning a separate page_object for each of our Application pages (at-least the ones we are testing), grouping reusable pieces like our breadcrumb and other app globals under spec/page_objects/sections and definning page specific HTML content to be under the respective page object.

## Tests

## Types of changes

## Checklist:


spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -63,6 +64,7 @@
 
   scenario 'Case worker creates and deletes assessment' do
     login
+    @assessment_changelog = AssessmentChangeLog.new
```
zliuwd: Go through all of these four scenarios, only here is not a "step", should we define this constant in another where or wrap it up with other "steps? After this, all scenario "steps" will be more clear.

spec/page_objects/sections/app_globals.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AppGlobals < SitePrism::Section
```
zliuwd: For here we separate some elements from AssessmentGlobal to AppGlobal, IMG，the basis of this division will be one of these two down below:
1. where they are defined
2. where they are invoked
Based on this, global alert and modal all are defined in component/common, and all are invoked in assessment form page, so I think that separating them in two "global" files still not very clear.
Additionally, if we look at how they are include in AssessmentFrom page (as below), we will found both these two sections are point to same html element "body", this also proved that this division of these two section is not very clear.
```
class AssessmentForm < SitePrism::Page
  section :app_globals, AppGlobals, 'body'
  section :global, AssessmentGlobal, 'body'
```

PR #455, opened 2019-01-24
Add a release toggle for CANS 2.0

wmitchell-cambria:
## JIRA

Pre-requisite for [CANS-946 Replace React Wood Duck headers with Design System headers from Component Library](https://osi-cwds.atlassian.net/browse/CANS-946)

## Description

This adds the ability to flag CANS 2.0 functionality from an env variable. More broadly, it also introduces the same approach that Intake uses to pass environment variables down to the JavaScript via the window object, so that future feature toggles can be created more easily.

Comments regarding the use of `window` here are valid... This primary benefit of this approach right now is its consistency with other services. It is known to work well, and the list of active features is available from the moment the page loads.

This PR should also unblock several other stories that will require [Feature Toggles](https://github.com/ca-cwds/development-practices/blob/master/feature_toggles.md) this sprint.

Notes:

* To turn on CANS2.0 functionality, add this to your .env file: `CANS2=true`
* Generally, feature toggles should be more granular than this. In trade, this one should be very short-lived.

## Tests

## Types of changes

## Checklist:


wmitchell-cambria:
Retest this please

app/javascript/Application/util/features.test.js
```diff
@@ -0,0 +1,39 @@
...
+    })
+  })
+
+  describe('isCans2', () => {
+    it('is true if the feature is in window config', () => {
+      window.org = Object.freeze({ cans: { active_features: ['Feature One', 'cans2'] } })
+      expect(isCans2()).toBe(true)
+    })
+
+    it('is is false if the feature is not in window config', () => {
```
pokharelmanish: typo: is is?

app/javascript/Application/util/features.test.js
```diff
@@ -0,0 +1,39 @@
...
+    beforeEach(() => {
+      window.org = Object.freeze({ cans: { active_features: ['Feature One', 'Feature Two'] } })
+    })
+
+    it('is true if the feature is in window config', () => {
+      expect(isFeatureActive('Feature One')).toBe(true)
+      expect(isFeatureActive('Feature Two')).toBe(true)
+    })
+
+    it('is is false if the feature is not in window config', () => {
```
pokharelmanish: is is ?

wmitchell-cambria: It really is is. (Good catch)

spec/acceptance/index_page_spec.rb
```diff
@@ -27,4 +27,13 @@
...
   end
+
+  scenario 'CANS pages have active feature information' do
+    login
+    active_features = page.evaluate_script(
+      'JSON.stringify(window.org.cans.active_features)'
+    )
+    expect(active_features).to start_with('[')
+    expect(active_features).to end_with(']')
+  end
```
AK56: Let's move this scenario into our Regression suite

wmitchell-cambria: Why? Eventually they will be merged into one suite, and having it here ensures that it runs earlier in the development process (PR builds). It doesn't depend on any data or specific credentials, either.

AK56: Some thoughts: We are going to have a divergence in growth of our testing suites. I feel like we have quite of a robust regression suite and should grow that one instead. It can run against Int or PreInt env controlled by an env variable. We have login helpers in place and perhaps should replace or add in the regression run in pipeline for PR builds. I'm happy to chat more offline on this.

wmitchell-cambria:
Invitation sent for the teams to sync up on Tuesday, so that everybody can understand how to use Feature Toggles. In the meantime, I would not like to wait on this PR, since it blocks my current story.

spec/acceptance/index_page_spec.rb
edgibbs: Coming from experiences with using feature toggles and testing, you want to go really minimal in test suites that depend on temporary feature toggles.  Generally just go ahead and test with the new features always turned on.  If you think a feature flag is going to last for a long time maybe it's worth a single test to see that you can't access it if the flag is turned on.

edgibbs: And we're talking about Release Feature toggles here where the feature is turned off for everyone.

wmitchell-cambria: True. Re: Gurpartap's comments on the location of this test, I think it may actually be best served being merged in as an expectation in a handful of other tests, rather than its own. I primarily wanted to add it here to ensure that the features were being propagated to JavaScript at all. We may end up needing test suites for particular release toggles, but they should definitely also be cleaned up once the feature is released and the toggles are removed.

And good point re: Rails controllers. I think the same flags will be accessible there using `Feature.active_features` (but maybe we should consider abstracting decision points more than that).

wmitchell-cambria: I'm going to merge this as-is and follow up with a test refactor if there is time. (Remind me if I don't, please.)

LeonidMarushevskyi:
@wmitchell-cambria I will express my questions and concerns here and you can address them as soon as you have time or during our meetings on Tuesday.

CANS already has a set of application-specific permissions (Permission Toggle in terms of your document) which we are using in 1.1. Permissions are provided by Perry/IDM and not hardcoded as environment variables. Makes application dynamic and managed by admins from counties. In essence, permissions already work as feature toggles. Why we don't want to use permission approach to manage access to new features of CANS 2.0? In my opinion current CANS permission model solves the need for potential continuous delivery to production in the future and bring benefits of more dynemic feature management on user level instead of environment level. To cover partially developed features, which seems like the main reason for Release Toggle, we can use some more generic environment variable like DEV-IN-PROGRESS which should not leave longer then 1-2 sprints while feature is in developemnt. DEV-IN-PROGRESS can be also provided by Perry instead of hardcoded as environment variable. Toggles in Perry gives much more flexibility in testing as well.

The notion of Release Toggle CANS2 as environment variable brings issues. Let's imagine we will have CANS2 enabled in PreInt and Int always and disable it in Staging and Prod. It means all automated and manual testing on PreInt and Int will happen for code 2.0. Then we have a need for hotfix in 1.1 and release to production. As I understand you want to use the most recent code, fix bug in this code and deploy to Staging with CANS2=false parameter. With such scenario we are running a risks of production regression introduced by new development of 2.0 because application with CANS2=false is not a part of the development pipeline and never tested as a part of development process. It means we need to test application with all permutation of feature flags and it means redeploy and much longer pipeline build time, additional work building this type of pipeline, duplication of tests and much more things. We went through this with Perry where we now already testing two different modes.

Another aspect is that in the scope of CANS 2.0 we are making a lot of changes to existing 1.1 features making them more mature. If we will try to introduce conditional logic base on release toggle all over the place trying to maintain two branches of code - it will definitely cause issues with code quality and complexity of the code, slow down the development, increase testing overhead. Then additional work to remove this conditional logic when release happened.

While technically speaking we already capable to automate CANS delivery up to Production (we have automated regression tests both in Int and Staging as well as an automated performance test suite, security scan for Docker images), we can't anticipate that organization will switch to continuous delivery in nearest future.

The organization established in a way which is not expecting continuous delivery to production (traditional government project approach). I don't see how in next several months we will be able to change delivery model bypassing existing processes with manual QA testing in Integration, user acceptance testing in Staging, implementation cycles, security scanning, etc. Releases will continue to happen each 2 to 3 month. If no continuous delivery planned yet why to take the technical overhead of feature toggles now?

With this said, it is much safer (at least for CANS 2.0) to use hotfix approach discussed earlier with John O'Duinn and John Simko where we supposed to release hotfixes, deploy them to PreProd and test there (https://github.com/ca-cwds/cws-cares/blob/master/images/DevOps-Pipeline.png). We don't need any long leaving branches for this and it is not opposed to trunk based development approach. We have tags for each version and we can alway branch from release tag, fix the bug in this branch, release new hotfix tag and build application from it. Fixes should also be applied to master branch. We used this approach from the project beginning on all modules like Dora, Perry, CALS API, etc and it works well. With this approach there is no risk to introduce Production regression because we make very small changes to production code.   

CANS2 does not seem like a short living toggle according to what I described above. We can anticipate CANS 2.0 release around March-April. 

I propose to use Permissions for new features instead of the environment variable with the release version. We can expand list of CANS specific permissions according to new functionality. Continue to use hotfix flow until the organization will make the decision to introduce continious delivery up to production.

In case of CANS application, implementation of feature flags according to https://martinfowler.com/articles/feature-toggles.html will bring significant development overhead, technical risks while not bringing much value without actual continuous delivery in place.  

I'm not trying to say that the general idea of feature flags is bad. We already partially using this approach where it is needed (permissions, configurations, modularity). We just need to be careful before introducing unneeded complexity in each particular case. In case of CANS 2.0 introduction of such approach looks like overengineering and can increase delivery time. More thoughts about feature flags here https://dzone.com/articles/feature-toggles-are-one-worst

edgibbs:
I think we're all on the same page that you want to use feature flags as little as possible, but if we're ever going to start delivering at an incremental pace we'll have to have a few.  I think we can avoid using release toggles most of the time by favoring:

- Not using toggles at all, say you modify labels on a form, you just release it.  Or you add a new column to a report the users can just access it.  If you're doing small releases this should be fine.
- Not implementing the linking of the feature into the UI so if you happen to know the URL path you can access it, but if not it's basically invisible.
- Using permission toggles with a permission no one has been granted.
- Not building the UI part of a story so it's just a technical task at that point
- Release toggles if none of those suffice.

wmitchell-cambria:
Permission Toggles and Feature Toggles serve two different purposes. To support trunk-based development, we need both.

The `cans2` release toggle is likely poorly named and not described well enough. I’ll take the blame for that. It should not last until March/April. Instead, I only intended it to be used over the next week or two, until the code from Staging has progressed to Prod. Once that phase is over, we can remove it in favor of granular Feature Toggles for individual use cases. (Or better yet, developing in a way as Ed described where we feel comfortable pushing many things out without Toggles.) I’ll describe a little more on how we can use this _particular_ flag in the meeting tomorrow, and what the timeline is for its removal.

In this situation, a Permission Toggle is not appropriate. Over the next couple weeks, we want to ensure that no CANS 2.0 functionality is accidentally released to Production. Could we do this with a user-based permission? Yes. However, that would involve modifying many individual users in Integration. When it comes time to allow CANS 2.0 in Staging, we would need to do the same there, modifying a whole list of users instead of turning on a single environment flag. Permission Toggles are best served for cases where the flag is much longer-term, and more dynamic. In this case, applying a configuration to the entire environment saves a lot of overhead.

I’ll go over the process for “hot fixes” during tomorrow’s meeting. You raise some good points, although there are challenges with the alternate approach as well. The topic of Feature Toggles primarily arose _after_ that diagram was created, when we ran into difficulty implementing it in the aftermath of the Snapshot release. In other words, this is not over-engineering based on imagined pain, but a response to real challenges. Many of our pipelines are set up only for trunk-based development, not a branching approach, especially when considering dependencies between applications. (One example…with the current PR to use Regression Tests during PR builds, any CANS branch built must be compatible with the HEAD of CANS-API, not a release tag.)

Regarding overhead, technical debt, and increased complexity, nobody is saying that we should create a nest of long-lived feature toggles. Toggles introduce complexity and should be managed carefully. The Martin Fowler article (as well as the article you mentioned) both give lots of advice on managing them (recognize their cost, minimize their entry points, etc). We should not go overboard on very granular Feature Toggles… but we do need more than zero.

If we can use these to support continuous delivery through Staging and maintain the ability to “hot patch” Production, then that is a huge step in the direction of more frequent, incremental releases to Production. Otherwise, we are stuck in a Catch-22, where we cannot deploy frequently and safely to Production without proper infrastructure, nor will we create proper infrastructure unless we are deploying frequently to Production.

Lastly, I just need to call out here that there is not an approach here that has _no_ risk. Any release of code to production has some risk of regression. We are discussing multiple ways of reducing that risk, but there will _always_ be some amount remaining.

PR #457, opened 2019-01-25
CANS-909 Fix Caregiver Domain Shows Previous Domain's Name

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-909

## Description
The Caregiver Name from the last caregiver domain on the Assessment appears in the subsequently created caregiver domain when initially creating the new caregiver domain, if the user clicks an "Add Caregiver" link that is not on the last caregiver domain card.

When you Add a new caregiver domain now from the previously created domains in the list. It will create a new domain and not reflect the name from the previous domain. 

## Tests

## Types of changes

## Checklist:


nasar-cambria:
jenkins re-test this please

nasar-cambria:
re test this please

nasar-cambria:
Had an offline discussion with @wmitchell-cambria and addressed his questions.

wmitchell-cambria:
I can summarize part of the offline discussion. At the root, this is an issue with how we are generating the `key` property on Domains. The key includes the caregiver domain index (`a`, `b`, `c`, etc). When we add a new caregiver domain in the middle, we update all of the indices to be in alphabetical order again. So if you click `Add Caregiver` on domain `b`, then the new domain will become `c`, the old `c` will become `d`, `d` will become `e`, and so forth. Because of this shift, React gets confused and thinks that the new caregiver is actually an update of the old one. So, it does not reset the internal state, which includes the caregiver name.

I agree with the Request Changes on this one. We should fix the root of the issue, by creating a different unique key that can track domains as they are reordered. (Otherwise, we will have similar state-tracking problems as we delete caregiver domains.)

nasar-cambria:
@wmitchell-cambria As discussed, we are now using a special id to track caregiver domains which will result in a unique key every-time we add or delete caregiver domains. By doing this we are able to correctly  track domains as they are re-ordered. 

@denys-davydov this doesn't impact collapse / expand domains functionality.

wmitchell-cambria:
@nasar-cambria - Did you confirm that using and _saving_ the `id` in this way is acceptable within our data model?

nasar-cambria:
@wmitchell-cambria I checked with Leonid and he confirmed that it is okay. Since it will be save in the Json and gets submitted once.  I also verified if we have any functionality related to the order the ids but found none. So, I think we are good to go.

wmitchell-cambria:
I guess my final comment here is that the test coverage is probably a little weak, given how deep we had to dig into the problem. Can you just double check the coverage levels before and after and make sure they look OK?

wmitchell-cambria:
kk, currently master for that file is at 97.7 | 90.63 | 92.31 | 98.8 |. So, that's mostly a decrease. Something to watch out for.

nasar-cambria:
@abelayt sure, request noted. We can pair up sometime today or tomorrow and I can show you how this is working.

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -161,12 +169,12 @@ class Assessment extends Component {
...
             <CardBody>
               {domains.map((domain, index) => {
-                const { code, caregiver_index: caregiverIndex } = domain
+                const { id, code, caregiver_index: caregiverIndex } = domain
                 const domainI18n = getI18nByCode(i18n, code)
                 return (
                   <Domain
                     index={index}
-                    key={code + caregiverIndex + isDefaultExpanded}
+                    key={code + caregiverIndex + isDefaultExpanded + id}
```
denys-davydov: Please try without _caregiverIndex_ here

nasar-cambria: removed.

PR #461, opened 2019-01-28
CANS-964 - Remove/Add Caregiver links

abelayt:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/style.sass
```diff
@@ -196,20 +196,17 @@ $item-confidential-label-height: 6.05rem
...
-  color: $aspen
-  cursor: pointer
-  font-weight: bold
+.caregiver-domain-add-remove
+      float: right
+      margin-top: 2.0rem
+      margin-right: 1.5rem
+
+.caregiver-domain-add-remove-links
+      font-size: 1.4rem
```
denys-davydov: "ADD CANS" button on the Client Profile page has a size of 12.5px. AC #4 states that "The text links are designed to match the font size of the "Add CANS" text link on the client profile"

app/javascript/Application/components/Assessment/DomainCaregiverControls.js
```diff
@@ -1,39 +1,34 @@
...
-              aria-label="add caregiver button"
-              tabIndex={0}
-            >
-              + ADD CAREGIVER
-            </div>
-          </li>
-        </ul>
-      </h5>
-    </div>
+    <Fragment>
```
denys-davydov: There is no need in Fragment if it has only one child, like here it has only a div as a child

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -196,20 +196,17 @@ $item-confidential-label-height: 6.05rem
...
-  font-weight: bold
+.caregiver-domain-add-remove
+      float: right
+      margin-top: 2.0rem
+      margin-right: 1.5rem
+
+.caregiver-domain-add-remove-links
+      font-size: 1.25rem
+      font-weight: bold;
+      color: $light-blue
```
wmitchell-cambria: Are these styles necessary? When using `<Button color="link"/>` from the component library, I would expect the component library to provide styles for font-weight, color, and text-decoration.

If we are overriding component library styles for this particular situation, we should consider checking with DesignOps to see if that is really necessary. The component library should be as close to a drop-in solution as possible.

abelayt: I checked with Robert from DesignOps and showed him the links before I changed the font-sizes as Denys mentioned from the A/C (from 1.4rem to 1.25rem). Robert approved and said they are still working on the global design for CWS-CARES  and these and many others will be changed. Your point, we should not override component library styles, should still be the case even now with what we have in the component library, but that is not the case with some of the styles as of now.

wmitchell-cambria: If these styles will change, it seems even more important to use the default component library styles. That way, when the global styles are updated, CANS will not lag behind.

abelayt: The default component library styles does not meet the A/C for this story. That is the reason we need to style to match what we have globally.

abelayt:
retest this please

PR #464, opened 2019-01-30
CANS-966-Summary Card Available before Assessment Completes

AK56:
Please checkout the story for details: https://osi-cwds.atlassian.net/browse/CANS-966

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -287,6 +290,16 @@ class AssessmentContainer extends Component {
     }
     if (this.state.assessment.id) {
       this.updateIsEditableState()
+      const canDisplaySummary =
```
wmitchell-cambria: This appears right after a call to `this.updateIsEditableState()`, which also sets the `canDisplaySummary` state using different logic. Are these both necessary? A similar thing happens elsewhere in the file. It makes it really difficult to follow when the summary is being shown, and it adds a lot of complexity to an already large component.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -287,6 +290,16 @@ class AssessmentContainer extends Component {
...
+        this.state.assessment.status === AssessmentStatus.inProgress &&
+        this.state.isEditable &&
+        this.state.isValidForSubmit
+      this.setState({ canDisplaySummary })
+      if (this.state.canDisplaySummary) {
```
wmitchell-cambria: Note that `this.setState` is asynchronous, so this will operate on the previous value, not the newly set value.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -287,6 +290,16 @@ class AssessmentContainer extends Component {
...
+        // scroll the page upwards when assessment summary can display on save
+        const positionAdjust = -25 // for manually adjust scroll destination -25 means go up 25px more
```
wmitchell-cambria: Magic number. I feel like I have seen this one before... That probably means that the scrolling behavior is copied between two places?

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -413,7 +428,7 @@ class AssessmentContainer extends Component {
           />
         </div>
         <AssessmentSummaryCard
-          assessmentStatus={assessment.status}
+          isSummaryAvailable={canDisplaySummary}
```
wmitchell-cambria: One suggestion that may make this all a lot cleaner is to factor it out like so.

First, note that the assessment summary is only displayed if the assessment is complete OR if the assessment is both filled and saved.

1. In AssessmentContainer, track only the `hasBeenSaved` property in state, using minimal logic.
2. Create a new wrapper component for AssessmentSummaryCard, and pass it all of the props it needs to decide whether to render the component (`hasBeenSaved`, `status`, etc).
3. Within that new component, use those props to conditionally render the summary.
4. Move the scroll behavior into that component or the summary card itself. (Actually, probably wrap the summary in a new ScrollOnMount component.) Every time ScrollOnMount mounts, scroll to it.

With this approach, AssessmentContainer doesn't get too much bigger, the logic is consolidated into one place, and the scroll behavior is consolidated into one (separate, reusable) place.

AK56: Yes I can agree with the add on value by moving parts out of AssessmentContainer component, let's chat offline to solidify the approach

PR #470, opened 2019-01-30
Code 140 Refactor Assessment Container

vikas199:
## JIRA
CODE-140(https://osi-cwds.atlassian.net/browse/CODE-140)
## Description
This story is about refactoring the assessment container and code climate fix.

## Tests

- [x] I have included other tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -57,7 +52,10 @@ class AssessmentContainer extends Component {
   }
 
   async componentDidMount() {
-    this.initAssessment().then(() => this.updateIsEditableState())
+    const assessmentId = this.props.match.params.id
+    assessmentId
+      ? this.fetchAssessment(assessmentId)
```
wmitchell-cambria: Shouldn't this have a `then` as well? (Little things like this would make me worry that the refactor won't behave exactly the same as it did before.)

app/javascript/Application/components/Assessment/RenderAssessmentContainer.js
```diff
@@ -0,0 +1,94 @@
...
+import Assessment from './Assessment'
+import AssessmentSummaryCard from './AssessmentSummary/AssessmentSummaryCard'
+import { LoadingState, isReadyForAction } from '../../util/loadingHelper'
+import { AssessmentFormHeader, AssessmentFormFooter } from './'
+import { completeTip } from './AssessmentHelper'
+import RenderWarning from '../common/RenderWarning'
+import WarningShow from '../common/WarningShow'
+import { isCompleteAssessmentAuthorized } from '../common/AuthHelper'
+
+const RenderAssessmentContainer = ({
```
wmitchell-cambria: I appreciate that you are trying to split things out into their own components, but I have concerns about how they are being split.

1. This component has over 20 props. That's still quite a lot!
2. There is no better name for this tahn `RenderAssessmentContainer`
3. Most of the logic needed to generate these props still resides in AssessmentContainer itself, so that component may be shorter, but not really much less complex.

Instead, we should try to find ways of separating out chunks of behavior into their own components. Individual components should have:
* One responsibility, and a good name that describes it
* A handful of props at most
* Good test coverage

app/javascript/Application/components/Assessment/AssessmentHelper.js
```diff
@@ -184,3 +189,17 @@ export const completeTip = (
     The Assessment Date and all assessment ratings must be completed before the Complete button becomes active.
   </Typography>
 )
+
+export const getCaregiverDomainsNumber = domains => {
```
wmitchell-cambria: This seems like a reasonable helper method (although it is only used in one spot). The others, not so much. They either do things that only the AssessmentContainer needs (alertMessage, updateUrlWithAssessment) or are named in ways specific to their situation (alertMessage, handleCountyName).

Due to this, it is not really factoring out reusable pieces of code, but rather moving it to get around the file limit. Somebody looking at AssessmentContainer still has to consider just as much complexity, but now it is split between two files.

app/javascript/Application/components/common/RenderWarning.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { caregiverWarning } from '../Assessment/AssessmentHelper'
+import PropTypes from 'prop-types'
+import PageModal from '../common/PageModal'
+
+const RenderWarning = ({ isCaregiverWarningShown, handleWarningShow, handleCaregiverRemove, focusedCaregiverId }) => {
+  return (
+    <div>
+      {isCaregiverWarningShown ? (
```
wmitchell-cambria: These RenderWarning and WarningShow components are the exception to my comment above. I do think that these are good ways to break out the functionality, although they need better names.

Note that PageModal should have an `isOpen` prop that when set to false renders nothing. You can use that instead of the ternary here. This is often a good way of separating out functionality:  move the logic from the parent into the child by passing the child everything it needs to know (isCaregiverWarningShown in this case) and removing conditional logic from the parent.

app/javascript/Application/components/Assessment/AssessmentContainer.js
vikas199: there is bug in this method I'm making a fix

app/javascript/Application/components/Assessment/RenderAssessmentContainer.js
vikas199: Test coverage on master:    96.87 |    92.35 |    93.45 |    97.11 |  
Test coverage on Branch:    97.15 |    92.35 |    93.77 |     97.4 |

app/javascript/Application/components/common/RenderWarning.js
vikas199: removed the itenary logic and gave it to the Page modal

app/javascript/Application/components/Assessment/RenderAssessmentContainer.js
vikas199: changed the name from RenderAssessmentContainer to AssessmentContainerInnder

app/javascript/Application/components/Assessment/AssessmentContainerInner.test.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+describe('AssessmentContainerInner />', () => {
+  describe('renders component and shows warning', () => {
+    it('renders with 1 <AssessmentSummaryCard /> component', () => {
+      const wrapper = shallow(<AssessmentContainerInner {...props} />)
+      expect(getLength(wrapper, AssessmentSummaryCard)).toBe(1)
+    })
+
+    it('renders without <AssessmentFormHeader /> component initially', () => {
```
zliuwd: Typo "without"?  Should be `renders with <AssessmentFormHeader /> component initially` ?

vikas199: Done.

vikas199:
jenkins retest this please

zliuwd:
Looks good to me, decent testing coverage, even though I don't think AssessmentContainerInner is the best name. Couple small suggestions, nonblocking.

PR #477, opened 2019-01-31
Revert "CANS-923: Update webpack-dev-server dependency"

LeonidMarushevskyi:
Reverts ca-cwds/cans#463 due to issue in PreInt. Tests failed on Preint http://jenkins.mgmt.cwds.io:8080/job/PreInt-Integration/job/deploy-cans/78/Regression_20Tests_20Dev/ and application is down.

Error in the log on PreInt 
Webpacker can't find application.css in /app/public/packs/manifest.json. Possible causes:
1. You want to set webpacker.yml value of compile to true for your environment
2. webpack has not yet re-run to reflect updates. (edited) 
3. You have misconfigured Webpacker's config/webpacker.yml file.
4. Your webpack configuration is not creating a manifest.


LeonidMarushevskyi:
retest this please

wmitchell-cambria:
@LeonidMarushevskyi - What error are you seeing that this seems to be causing? I see the Preint build failing for several changes _before_ this was merged. Builds 75 and 77 have a regression test error related to change log, and builds 76 and 78 have regression test errors around creating an assessment.

This change could definitely cause issues... and we should be careful about it. I think I'm just missing something.

wmitchell-cambria:
kk, chatted with Leonid offline. Seems like this is another issue in addition to the Regression failures.

PR #485, opened 2019-02-04
CANS-665 Domain or Item scroll up when it is expanded

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-665

## Description

After user expanded a domain or an item,  its header will automatically scroll to the top part of page.

## Rmarks
Finally, have to use sleep in regression test to eliminate the impact that autoscroll brought by (after tried multiple waiting approaching includes some methods from lib or from pure ruby code, all did not work ). The test only could get right result after autoscroll stop, so we have to have a "hard waiting" after that.
## Tests

## Types of changes

## Checklist:


zliuwd:
retest this please

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -180,6 +180,7 @@ def input_date_and_calendar_icon_test(current_date)
   def domain_and_item_rating_test
     @form.collapsed_domain_headers[0].click
     expect(@form).to have_domain_level_reg_rating
+    @form.domain_level_reg_rating[0].click
```
zliuwd: I have to add on this line to resolve the racing between "autoScroll behavior" and  "capybara click".

zliuwd:
retest this please

zliuwd:
retest this please

zliuwd:
retest this please

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -89,6 +87,23 @@
     save_and_check_the_success_message
   end
 
+  def eliminate_auto_scroll_impact
+    sleep 2
+    @form.item_comment_icons[0].click
```
denys-davydov: Not sure I understand this line in the scope of _eliminate_auto_scroll_impact_ function. Please elaborate.

LeonidMarushevskyi: Scroll makes impossible to click the icons. So we are waiting until scroll will finish and then clicking it.

zliuwd: Hi, @denys-davydov  , thanks for review. Capybara' s default waiting time coincide with the auto scroll in this situation (because we used setTimeout to trigger the scroll), then the test will fail to click. So we add on sleep to wait for autoscroll finishing, then we click a "blank place" here is that comment icon to make sure the page get focused again. If we don't have that "re-focused" step ,some time, we will miss next click.

denys-davydov: So the purpose of this function was to click the icon? Then it is not the best naming here.

zliuwd: Thank you Denys 👍

PR #487, opened 2019-02-04
Upgrade react-widgets dependency version

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-921

## Description
Restore CANS react-widgets dependency to published version

## Tests

## Types of changes

## Checklist:


denys-davydov:
The a11y test suite is failing with the following error:
`1) Pages are accessible Assessment Form page is accessible
     Failure/Error: expect(page).to be_accessible.excluding '#cancel-assessment'

       Found 1 accessibility violation:

       1) aria-valid-attr-value: ARIA attributes must conform to valid values (critical)
           https://dequeuniversity.com/rules/axe/3.0/aria-valid-attr-value?application=axeAPI
           The following 1 node violate this rule:
           
               Selector: #assessment-date_input
               HTML: <input id="assessment-date_input" role="combobox" placeholder="mm/dd/yyyy" aria-haspopup="true" aria-labelledby="assessment-date-label" aria-expanded="false" aria-owns="" type="text" tabindex="0" autocomplete="off" aria-disabled="false" class="rw-widget-input rw-input" value="02/04/2019">
               Fix all of the following:
               - Invalid ARIA attribute value: aria-owns=""`

It is despite the fact that a manual Axe Chrome Extension execution ran with no a11y violations. There is a difference in versions of **axe-core** that are used by _Axe Chrome Extension_ (3.1.2) and by _axe-matchers 2.1.1 gem_ (3.0.1). I've tried to upgrade _axe-matchers_ gem to 2.2.1 that has 3.1.1 version of axe-core
[package.json](https://github.com/dequelabs/axe-matchers/blob/11cacf42497b4ff9b103a286d72ea5484cd49689/package.json#L15), but it didn't help, not only it showed more violations. 

Here is a PR that adds allowEmpty option to all the _aria-xxx_ elements https://github.com/dequelabs/axe-core/pull/1154
It was merged on Nov 14th, 2018.
It sounds weird but axe-core 3.1.2 version was released 5 months ago, that is earlier than this PR. https://www.npmjs.com/package/axe-core/v/3.1.2 
However, the library has a lot of newer canary releases (like **3.1.2-canary.a34165c** that is the freshest as of now), so it looks like Axe Chrome Extension is using one of those canary releases. 

Proposed Solution:
1. Ignore the element in a11y tests for them to run green
2. Try to upgrade _axe-matchers_ gem to 2.2.1 and set it up to use **3.1.2-canary.a34165c** version of axe-core. If it is possible create PR and execute step 3. If not, wait for newer axe-matchers releases and proceed with a step 3.
3. Exclude the element from ignore

With the next commit, I will execute step one.

wmitchell-cambria:
That plan sounds great. If we end up delaying, we should just be sure to create a story so that we remember to come back to it.

spec/a11y/a11y_spec.rb
```diff
@@ -38,6 +38,6 @@
     find('#PSYCHOSIS-item-expand').click
 
     # this exclude should be discussed with designers
-    expect(page).to be_accessible.excluding '#cancel-assessment'
+    expect(page).to be_accessible.excluding '#cancel-assessment', '#assessment-date_input'
```
wmitchell-cambria: This is a silly request, but it is possible to put these on separate lines, like:

```ruby
expect(page).to be_accessible.excluding(
  '#cancel-assessment',
  '#assessment-date_input'
)
```

This will make it easier to view the git history if we add more exclusions. (Although, maybe not so much easier once we take commas into account.

denys-davydov: Like the approach and using it, but let's start to use it from 3 items :)

PR #489, opened 2019-02-04
FIT-368 add smoke tests to release pipeline

osi-richbach:
## JIRA

https://osi-cwds.atlassian.net/browse/FIT-368

## Description

Apply the new shared library for smoke testing to CANS


Jenkinsfile
```diff
@@ -327,6 +328,12 @@ def releaseToEnvironment(environment) {
   }
 }
 
+def smokeTestStage(environment) {
```
AVandana-OSI: @osi-richbach Don't we need to add any tests, instead of health check :)

osi-richbach: Debatable but I don't believe so...the story doesn't specify adding any tests....and my thought is really our job is to provide a facility for the tests not really implement tests...the knowledge of writing good tests rests more within the dev teams.  Plus, as you can see, the release pipeline already is running acceptance tests against preint and regression tests against integration.

osi-richbach: K, let's talk....this releasePipeline is already doing more than enough then I'm assuming with acceptance and regression tests at preint and integration respectively.

SergiiRedchuk: Common infrastructure for running smoke tests seems ready. A story for CANS team is needed to create or use existing smoke tests in a way that the tests will be executed be the newly added stage.

test/resources/smoketest.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+smoke=$(curl $1)
+
+if [[ !($smoke =~ false) && $smoke =~ CANS ]];
+then
+    echo "smoketest passed"
+else
+    echo "smoketest failed"
+fi
```
SergiiRedchuk: EOL is missing

osi-richbach:
went away from this and started a new branch

PR #492, opened 2019-02-06
Optimize and stabilize regression suite

denys-davydov:
## Description
This will improve the performance of login functionality and make 0-5 and 6-21 buttons clicks more stable.

## Tests

## Types of changes

## Checklist:

### DIFF
--- a/spec/acceptance_helper.rb
+++ b/spec/acceptance_helper.rb
@@ -27,14 +27,9 @@ def with_retry(func0, func1, times = 3)
       times = 0
     rescue StandardError => e
       times -= 1
-      if times <= 0
-        raise e
-      end
+      raise e if times <= 0
     end
-
-    if times <= 0
-      break
-    end
+    break if times <= 0
   end
 end
 
--- a/spec/acceptance_helper.rb
+++ b/spec/acceptance_helper.rb
@@ -18,6 +18,26 @@ STAFF_NAME = 'Regression, QA02'
 CLIENT_LIST_TITLE = 'Client List'
 CAREGIVER_DOMAIN_WARNING_MESSAGE = 'You are about to remove the CAREGIVER from this Assessment.'
 
+def with_retry(func0, func1, times = 3)
+  loop do
+    begin
+      func0.call
+      sleep(1)
+      func1.call
+      times = 0
+    rescue StandardError => e
+      times -= 1
+      if times <= 0
+        raise e
+      end
+    end
+
+    if times <= 0
+      break
+    end
+  end
+end
+
 def acceptance_helper
   return LoginHelper unless ENV.fetch('PROD_LOGIN', false)
   ProdLoginHelper
--- a/spec/acceptance_helpers/login_helper.rb
+++ b/spec/acceptance_helpers/login_helper.rb
@@ -9,7 +9,9 @@ module LoginHelper
   end
 
   def enter_credentials(login_config = default_json)
-    js_script = 'arguments[0].focus(); arguments[0].setAttribute("value", arguments[1]); arguments[0].blur();'
+    js_script = 'arguments[0].focus(); \
+                 arguments[0].setAttribute("value", arguments[1]); \
+                 arguments[0].blur();'
     input_field = find('input#username')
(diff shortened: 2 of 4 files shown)


spec/acceptance_helper.rb
```diff
@@ -18,6 +18,26 @@
 CLIENT_LIST_TITLE = 'Client List'
 CAREGIVER_DOMAIN_WARNING_MESSAGE = 'You are about to remove the CAREGIVER from this Assessment.'
 
+def with_retry(func0, func1, times = 3)
```
denys-davydov: The function takes 2 other functions (by design: action and assertion) and calls them in series. If the assertion fails it will try to repeat the scenario 3 times until success. With 3 fails in a row, the last exception will be raised.
When no exceptions, returns control.

spec/acceptance_helpers/login_helper.rb
```diff
@@ -9,7 +9,9 @@ def login(login_config = default_json)
   end
 
   def enter_credentials(login_config = default_json)
-    fill_in 'Authorization JSON', with: JSON.generate(login_config)
+    js_script = 'arguments[0].focus(); arguments[0].setAttribute("value", arguments[1]); arguments[0].blur();'
+    input_field = find('input#username')
+    page.execute_script(js_script, input_field.native, JSON.generate(login_config))
```
denys-davydov: This improves the performance of the login json input, puts the json immediately.

spec/acceptance_helpers/login_helper.rb
```diff
@@ -26,7 +28,7 @@ def click_logout
   private
 
   def need_login?
-    ENV.fetch('CANS_AUTHORIZATION_ENABLED', true) && !page.has_content?('CANS')
+    ENV.fetch('CANS_AUTHORIZATION_ENABLED', true) && page.has_css?('input#username', wait: 5)
```
denys-davydov: This improves the performance of the login operation, immediately returns true when on the login perry page. Saves us more than 30 seconds on every login.

spec/page_objects/assessment_form.rb
```diff
@@ -109,7 +111,13 @@ def check_case_or_referral_number
 end
 
 def click_0_to_5_button
-  @form.header.age_0_to_5_button.click
+  with_retry(Proc.new { @form.header.age_0_to_5_button.click },
```
denys-davydov: This will assure that the button is clicked with no glitches. It will repeat click up to 3 times if the button glitches.

spec/acceptance_helpers/login_helper.rb
wmitchell-cambria: This is great. My only concern is that there could be other pages other than the login page that have an input with ID username. Is there a way to continue using the "Authorization JSON" label to find it, or something else very specific to the login page?

spec/page_objects/assessment_form.rb
wmitchell-cambria: Can we create a tech debt story to investigate why the button is flakey and fix it? There was some DesignOps chatter yesterday about replacing it with a radio button. I wonder if that will help. `with_retry` seems like a good workaround for now, but ideally Capybara wait times should be sufficient in all cases. If it is not, we are probably causing unnecessary re-renders that interfere with the interactions, or something similar.

spec/acceptance_helpers/login_helper.rb
denys-davydov: We don't have and don't plan to have pages that will have input fields with "username" id. Authentication is fully on Perry side for CANS. Also, the described page has to be a home page "/" to break this logic.

spec/page_objects/assessment_form.rb
denys-davydov: Capybara wait time doesn't help here because the glitch behavior is following: the button is clicked, made selected, made unselected again. We are waiting 30 secs and it is still unselected. So for now, this workaround is the only way to stabilize I found. We can work more on it.

denys-davydov:
Retest this please

denys-davydov:
Retest this please

PR #493, opened 2019-02-06
CANS-954-Adding more valid date formats for assesssment date

AK56:
https://osi-cwds.atlassian.net/browse/CANS-954

## Description
Users might format dates in few different ways when typing in a date and we have a pre-defined set of allowed formats now which we pass in to our react-widgets DateTimePicker through `parse` prop, which does internal validation and if user enters one of the allowed formats, it does the parsing into 'MM/DD/YYYY'

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/DateField.js
```diff
@@ -39,16 +39,15 @@ class DateField extends Component {
   }
 
   handleOnBlur = event => {
-    this.props.onRawValueUpdate(event)
-    const date = event.target.value
+    const date = new Date(event.target.value)
+    this.props.onRawValueUpdate({ target: { value: date } })
+    this.props.onChange(localToIsoDate(date))
```
wmitchell-cambria: Can you explain a little about the changes to this function? It looks like it passes a Date instead of a String to the parent callback, and we are calling `onChange` in every case now, instead of returning early. If these are bug fixes, we should add tests explaining the edge cases.

AK56: Ahhh, I don't think it needs to call parents `onChange`, this piece of code was already there in the function and I thought I removed it. 
It's passing a Date now because the internal validation of DateTimePicker seem to pass a Date object when `onChange` happens and which goes smooth through our external validations with moment. For `onBlur`, it passes a string, for which we would have to move around some logic in our `moment` validations. So, converting it to a Date obj instead saved me some grief.

wmitchell-cambria: kk, dates can be somewhat sensitive to changes like this, so if we could add a test case or two for blurring, I think that would be good. And please test thoroughly, especially sometime after 5pm (UTC midnight).

AK56: Unless I'm missing something, we do have a coverage for `onBlur` in place to check if it calls `onRawValueUpdate' passing the event with a Date obj instead of a string. At that point we should have had a test case with a string as well but I suppose we previously only allowed the format 'MM/DD/YYYY' which is what our `LOCAL_DATE_FORMAT`, the internal/external validations on date field worked.

PR #494, opened 2019-02-06
Fixing ProdLogin for RegressionSuite

AK56:
## Description

Reverting a change from https://github.com/ca-cwds/cans/pull/492
Our Regression Tests against Prod didn't run on PreInt-Int deployment because it used 'input#username' to check if login needed, but this CSS doesn't exist on Prod login pages, thus reverting to use the original but keeping the wait time to 5 sec, which was the original intent to keep the wait times low 
## Tests

## Types of changes

## Checklist:


spec/acceptance_helpers/login_helper.rb
```diff
@@ -30,7 +30,7 @@ def click_logout
   private
 
   def need_login?
-    ENV.fetch('CANS_AUTHORIZATION_ENABLED', true) && page.has_css?('input#username', wait: 5)
+    ENV.fetch('CANS_AUTHORIZATION_ENABLED', true) && !page.has_content?('CANS', wait: 5)
```
wmitchell-cambria: If `input#username` caused a problem here, won't it also cause a problem in `enter_credentials`, or am I missing something?

Also, note that the general idea here should not be to limit the wait time, but instead to wait on something appropriate. `!page.has_content` will always wait for the full length of time. Instead, `page.has_no_content` or `page.has_content('something on the login page')` will return as soon as their conditions are met. `wait: 5` is an OK workaround, but not great practice.

AK56: The problem it was causing is during the check of `input#username` CSS which only exists in our preint login page. Thus, the `need_login?` was returning false in Prod even though no login happened. So, it's better to use a CSS that is available in both our preint and prod (choosing CANS for now). `!page.has_content?` won't wait for full length since we are passing in a wait. 
Utilizing `page.has_no_content` or `page.has_content` would require login page to follow POM model, and I'm only pushing a quick fix for our Int deployment here.

wmitchell-cambria: Does Prod use this `need_login?` function but not the same `enter_credentials` function?

By full wait, I mean the configured wait time, whether the default or 5 seconds. `!page.has_content` will wait (in this case for 5 seconds) for the content to show up before returning, and `page.has_no_content` will return as soon as it does not see the element.

`page.has_no_content` and `page.has_content` do not require POM... we are already using `page.has_content`. Not sure what you mean by this.

wmitchell-cambria: Alternative to this approach:
* Leave this as-is, with the new `input#username`
* Override `needs_login?` in prod_login_helper to use the field that `prod_login_helper` needs. That way, both helpers have matching `needs_login?` and `enter_credentials` methods.

Since I have an OOP book next to me, I will also point out tangentially that this is a common issue with extending via inheritance.

AK56: Yes, Prod uses different `enter_credentials` defined in the `ProdLoginHelper`. I like the alternative approach but I feel it would desire some thought and time to make sure regression behaves on both envs, because we have quite much of shared logic for login at the moment. I would like to take this approach in a seperate PR and get this fix rolling for our Int deployment.

wmitchell-cambria: Sounds good to me. Can you make sure that the discussion happens?

AK56: Sure thing! I will put that in my todos..

PR #496, opened 2019-02-06
FIT-368 try 2 added smoke tests to release pipeline

osi-richbach:
## JIRA
FIT 368

## Description
Smoke tests stages are added to the release pipeline.

## Tests

## Types of changes

## Checklist:


osi-richbach:
retest this please

Jenkinsfile
```diff
@@ -341,3 +358,11 @@ def updateManifestStage(environment, version) {
     updateManifest("cans", environment, GITHUB_CREDENTIALS_ID, version)
   }
 }
+
+def smokeTestStage(environment) {
```
SergiiRedchuk: Looks weird that 'acceptance tests' or 'regression tests' are called here also as 'smoke tests'.

osi-richbach: Agreed on comment on shared library not being used here.  Although I'd debate on the benefit of using the shared library in it's current incarnation.  However, following the AC which calls for using the shared library, I won't get this finished tonite.  It will bleed into next sprint.

osi-richbach: as for the 'acceptance tests' and 'regression tests' also being called smoke, I tried to minimize refactoring everything all at once...In the end, there is quite a bit of confusion in there....acceptance tests are actually just regression tests but they are run against preint vs integration.

edgibbs: Yeah, smoke tests were a really early shared library and they only take a script name and then execute it, so not providing a lot of utility.  Probably worth a revisit to see if it makes sense to either update or remove as a shared library step.

osi-richbach:
@SergiiRedchuk based on discussion in standup( we are going to spike how to get more utility from the smoke test shared library and start discussion with CANS team on some refactoring ), can we get approval on this PR?

PR #499, opened 2019-02-07
CANS-991. Substance Use Disorder message alternates depending on age template

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-991

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentFormHeader/ConfidentialityAlert.js
```diff
@@ -1,16 +1,23 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { isEmpty } from '../../../util/common'
 
-const ConfidentialityAlert = ({ canReleaseInformation }) =>
-  canReleaseInformation ? null : (
+const ConfidentialityAlert = ({ canReleaseInformation, isUnderSix }) =>
+  !canReleaseInformation && !isEmpty(isUnderSix) ? (
```
wmitchell-cambria: Because `isUnderSix` will be defaulted to `null` when it is `undefined`, my personal preference would be to check `isUnderSix !== null` here in order to avoid the dependency on `util/common`. But it's just that... a personal preference.

denys-davydov: I am using a util function to cover an unexpected but a still possible situation when isUnderSix is `undefined`. Do we have any concerns with using our internal util functions?

wmitchell-cambria: No, no concerns with using the shared function. The only (minor) concern would be if the dependency graph gets complex, but I don't think that's something we need to worry about with this.

denys-davydov:
Retest this please

PR #507, opened 2019-02-11
CANS-946 - Toggleable Layouts for the rest of the pages

wmitchell-cambria:
## JIRA

Partial for [CANS-946 - Replace React Wood Duck headers with Design System headers from Component Library](https://osi-cwds.atlassian.net/browse/CANS-946)

## Description

This PR completes the 'page inversions' for the remaining page. That is, it refactors them to use the `FullWidthLayout` and load their own needed data explicitly. I tried to make pages which are largely similar (like Staff Changelog, Search Changelog, and regular Changelog) share as much code as possible. However, I think there is still more that can be done. I am pausing for review to help guide that continued refactoring.

~Apart from that continued refactoring, the next step for this story is to turn the DesignSystemLayout flag on in Preint and Integration.~ Cancel that. I think the next step will be to just remove the RWD layout in code, and deploy normally. (The flag was mostly helpful only for these smaller PRs.) Once we are happy with that, I can remove the ReactWoodDuck dependency, which will cause style breakages that need to be tracked down and fixed.

## Tests

## Types of changes

## Checklist:

I haven't run regression tests locally yet, actually. I'll see how those go.


wmitchell-cambria:
@AK56 - Good catch! That must have been a TODO that was lost between Friday and Monday. I've updated the AssessmentPage components to make sure that all info needed for the breadcrumb is being passed through.

wmitchell-cambria:
One side note... After these refactors, it should be possible to construct our breadcrumbs in a much more straightforward manner. Each page component can declare its breadcrumbs directly, and we can avoid many of the conditionals and complexity of the Pipeline.

wmitchell-cambria:
Yup. For future reference, the way that I went about debugging that was to check the URL, and then go to that page and see what params were being passed into the ContextualBreadCrumb. In this case, it was missing staffInfo, and adding that fixed it. But...yeah, some of that was based on my knowledge that staffInfo was supposed to be there. Otherwise I would have had to trace the Pipeline logic.

PR #511, opened 2019-02-13
CANS-988 Caregiver Name field required with red text prompt

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-988

## Description
Add a visible warning text to indicate caregiver name field is required and user cannot complete the assessment form without adding caregiver domain name.

## Tests

- [] I have included other tests

## Types of changes

## Checklist:


DmitriyRudenko:
both validateCaregiverDomainName and areItemsValid methods are iterating domains, which is not efficient

nasar-cambria:
Dimitry - it is iterating because caregiver domain need an explicit check to validate if it has care giver name. Also when ever user adds a new caregiver domain it need to iterate over all the domains and calculate if there is anything left unchecked or blank.. 

I am checking if is_caregiver_domain is true. That check needs to be there without it we cannot capture caregiver domain. 

When u are saying it is code duplication. Do you have an alternative approach?

PR #522, opened 2019-02-15
CANS-876-Collapse all domains when different age is selected

AK56:
Basically lifting up our `isDefaultExpanded` state so it can be passed in as prop and tweaked by our header and assessment components. 

## JIRA
https://osi-cwds.atlassian.net/browse/CANS-876

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainerInner.js
```diff
@@ -35,6 +36,13 @@ class AssessmentContainerInner extends Component {
     })
   }
 
+  handleExpandAllDomains = value => {
```
denys-davydov: Is the `value` parameter really needed here? Will it work if the function only toggles one boolean value from the state and has no parameter?
`handleExpandAllDomains = () => {	
    this.setState({	
      isDefaultExpanded: !this.state.isDefaultExpanded,	
    })	
  }`

wmitchell-cambria: I'm going to +1 this, and also piggy back to say:

`typeof value === typeof true` is weird. Can we just check that `value !== undefined`?

denys-davydov: @wmitchell-cambria yes, this line a little confused me too. I was thinking of `typeof value === 'boolean'` instead, and then I was like "Do we really need the `value` here?"

AK56: We are using the `value` param so our age selection can pass in a boolean value set the `isDefaultExpanded` state and keeping the ExpandAll/CollapseAll behavior intact which just toggles the `isDefaultExpanded` state. 
Also `typeof true` returns "boolean", which does the trick for us that `value === 'boolean'` or `value !== undefined` would have. I just happen to found it little more alluring

wmitchell-cambria: One of the reasons I like `value !== undefined` is because it clearly says that this value can be a boolean (since we set it to the isDefaultExpanded state) or undefined. With the `typeof` options, you know it can be a boolean, but it doesn't really explain what the default situation is. Is it undefined, or is it an event object, or something else?

In any case, I bet if you showed the `typeof value === typeof true` to some of our more junior devs, they would scratch their heads for a second. We should strive to make it as clear and readable as possible, even if the `typeof value === typeof true` is symmetrically aesthetic.

denys-davydov: @AK56 Thank you, now I got it. I'm fine with what we have then 👍

denys-davydov:
Retest this please

app/javascript/Application/components/Assessment/AssessmentContainerInner.js
AK56: @wmitchell-cambria I think it's fine for now, `typeof` shouldn't be that mind baffling for a JS developer plus it be a good opportunity to know about it.

PR #524, opened 2019-02-18
CANS-755 Refactoring item component

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-755

## Description

- Create <ItemToolbarControls/> component to hold N/A checkbox, confidential checkbox and rating radio buttons
- Refactored Item component and some related components by the way
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Item/ItemHelper.js
```diff
@@ -0,0 +1,122 @@
+import classNames from 'classnames'
+
+export const UNSET_RATING = -1
+export const NA_RATING = 8
+
+export const isNARating = rating => rating === NA_RATING
+
+/* eslint-disable camelcase */
```
wmitchell-cambria: Is this necessary? If it is, can we turn it back on after the offending lines?

app/javascript/Application/components/Assessment/Item/ItemHelper.js
```diff
@@ -0,0 +1,122 @@
...
+export const itemInnerComponents = Object.freeze({
+  ITEMHEADER: 'ITEMHEADER',
+  TOOLBARCONTROLS: 'TOOLBARCONTROLS',
+  ITEMDESCRIPTION: 'ITEMDESCRIPTION',
+})
+
+const getItemHeaderProps = all => {
```
wmitchell-cambria: I am skeptical of moving these 'prop dividers' to helper functions. These are not reusable functions...they only are useful in one component for a very specific set of properties. When reading or working on that component, the developer needs to come to look at this second file to understand what is going on. It would be more explicit if these were still included directly in the component file. If including them there makes the component too large, we should find other ways to simplify it.

app/javascript/Application/components/Assessment/Item/Item.test.js
```diff
@@ -452,3 +441,56 @@ describe('<Item />', () => {
     expect(wrapper.find('Comment').prop('disabled')).toBe(true)
   })
 })
+
+/* Down below is the test kit for the refactored Item Component
```
denys-davydov: I think these commented out lines should be removed. A good idea will be to have such a comment as a PR comment instead of a comment in a code base.

app/javascript/Application/components/Assessment/Item/ItemHelper.js
zliuwd: Ok, this is kind of attempt, the reason is just what you mentioned, the file size is too large before. I will use something like "ItemInner" to instead them, thank you!

app/javascript/Application/components/Assessment/Item/Item.test.js
zliuwd: Will do, thank you so much!

app/javascript/Application/components/Assessment/Item/ItemToolbarControls.test.js
```diff
@@ -0,0 +1,89 @@
...
+    expect(fakeProps.ratingType).toBe('REGULAR')
+    expect(target.exists()).toBe(true)
+    const expectedProps = ['itemCode', 'hasNaOption', 'rating', 'onRatingUpdate', 'disabled']
+    expect(Object.keys(target.props())).toEqual(expectedProps)
+  })
+
+  it('will render ItemBooleanRating  with correct props when ratingType NOT equal to REGULAR', () => {
+    const boolProps = { ...fakeProps }
+    boolProps.ratingType = 'BOOL'
+    const boolWrapper = mount(<ItemToolbarControls {...boolProps} />)
```
wmitchell-cambria: I suspect that the afterEach `unmount` is unnecessary...but if it is necessary, then we should make sure we also unmount when we create custom ones, too.

zliuwd: Good catch, added another unmount. Thank you!

PR #528, opened 2019-02-20
CANS-996: When user clicks breadcrumbs while on the assessment form modal appears to allow user to save before the routing

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/UnsavedDataWarning.js
```diff
@@ -36,13 +42,29 @@ class UnsavedDataWarning extends Component {
...
       this.close()
     })
   }
 
+  continueAction() {
+    if (this.state.event !== undefined) {
+      eventBus.post(this.state.event)
+    } else {
+      disableUnsavedValidation()
+      window.location.assign(this.state.href)
```
denys-davydov: Please use Redirect component to redirect

DmitriyRudenko: done

app/javascript/Application/components/common/UnsavedDataWarning.js
```diff
@@ -4,8 +4,16 @@ import { UNSAVED_ASSESSMENT_VALIDATION_EVENT } from '../../util/constants'
 import { eventBus } from '../../util/eventBus'
 import { Button, Modal, ModalBody, ModalFooter } from 'reactstrap'
 import { Icon } from '@cwds/components'
+import { breadCrumbOnClickHandler } from '../Layout/BreadCrumb/BreadCrumb'
+import { Redirect } from 'react-router'
+import { basePath } from '../../App'
```
wmitchell-cambria: Ditto here.

app/javascript/Application/components/common/UnsavedDataWarning.js
```diff
@@ -14,14 +22,18 @@ class UnsavedDataWarning extends Component {
     this.close = this.close.bind(this)
     this.onAction = this.onAction.bind(this)
     this.onValidation = this.onValidation.bind(this)
+    this.onBreadCrumbsClick = this.onBreadCrumbsClick.bind(this)
+    this.continueAction = this.continueAction.bind(this)
   }
 
   componentDidMount() {
+    breadCrumbOnClickHandler.onClick = this.onBreadCrumbsClick
```
wmitchell-cambria: !!

This mutates the properties of the breadCrumbOnClickHandler globally. That's pretty scary. Mutating objects in JavaScript like this can quickly lead to unmanageable code, where it's impossible to tell what one module has done to overwrite another. We need to find a different way of doing this.

A very similar approach would be to have the breadcrumb onclick fire an event, and have this modal listen for that event. That would decouple the two components, so that they both depend on a shared common module (the event bus). If that doesn't work, then we need to keep looking.

DmitriyRudenko:
Jenkins retest this please

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -103,6 +103,38 @@
     expect(@form.header.conducted_by.value).to eq(CONDUCTED_BY)
   end
 
+  scenario 'Case worker clicks breadcrumbs when assessment is unsaved then saves it' do
```
wmitchell-cambria: Do we need separate scenarios for these? They share a lot of boilerplate, and having too many regression tests repeating the same actions makes the entire suite slower and more difficult to run. Perhaps they could be combined into a single scenario (or even with an already existing scenario)?

DmitriyRudenko: yes, we have to test that original action actually happened(like breadcrumb redirect, assessment change log is shown, right page is displayed on browser "go back" button etc.

DmitriyRudenko:
Jenkins retest this please

PR #530, opened 2019-02-20
CANS-949 User can access the change log from assessment form

nasar-cambria:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-949

## Description
TO create an ability to access the change log while viewing the assessment.

## Tests

## Types of changes

## Checklist:


nasar-cambria:
jenkins re test this please

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Redirect } from 'react-router-dom'
+import { UNSAVED_ASSESSMENT_VALIDATION_EVENT, CHANGE_LOG_EVENT } from '../../util/constants'
+import { Button } from '@cwds/components'
+import { eventBus } from '../../util/eventBus'
+
+class AssessmentChangeLog extends React.Component {
```
wmitchell-cambria: Component name doesn't match file name here.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -20,6 +20,10 @@ $item-confidential-label-height: 6.05rem
 .authorization-warning
   margin-left: 0.5rem
 
+.view-changelog-link
+  margin-right: 74rem
```
wmitchell-cambria: That seems big. Are you sure it shouldn't be positioned from the left or relative to some other element?

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -20,6 +20,10 @@ $item-confidential-label-height: 6.05rem
...
+  cursor: pointer
```
wmitchell-cambria: Probably redundant. Why is our Button not already `cursor: pointer`?

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
nasar-cambria: good catch, Updated.

app/javascript/Application/components/Assessment/style.sass
nasar-cambria: It needs to be a part of the same row where other buttons are rendered so using margin.

app/javascript/Application/components/Assessment/style.sass
nasar-cambria: you are right. Removed!

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
```diff
@@ -0,0 +1,57 @@
...
+    if (shouldRedirect) {
+      const { assessmentId, assessmentStatus } = this.props
+      const pathname = `${assessmentId}/changelog/${assessmentStatus}`
+      const redirectProps = { pathname }
+
+      return <Redirect push to={redirectProps} />
+    }
+    return (
+      <Button className={'view-changelog-link'} color="link" onClick={this.onClick} onKeyPress={this.onClick}>
+        <strong>VIEW CHANGE LOG</strong>
```
wmitchell-cambria: nit: All caps should be done with CSS text-transform instead of the HTML itself.

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
```diff
@@ -0,0 +1,57 @@
...
+    )
+  }
+}
+
+AssessmentChangelogLink.propTypes = {
+  assessmentId: PropTypes.number,
+  assessmentStatus: PropTypes.string,
+}
+AssessmentChangelogLink.defaultProps = {
+  assessmentId: 1234,
```
wmitchell-cambria: This is probably not a useful defaultProp. It seems like assessmentId would be required for this component. (Actually, it seems like maybe it should be optional, because it will be null/undefined for _new_ unsaved assessments. In that case, there should probably be no link to the change log, correct?)

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -6,6 +6,7 @@
 require 'page_objects/client_profile'
 require 'page_objects/staff_dashboard'
 require 'page_objects/assessment_changelog'
+require 'byebug'
```
wmitchell-cambria: Why?

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -68,6 +69,38 @@
     verify_the_content_of_summary_card('6to21')
   end
 
+  scenario 'Case worker attempts to access changelog and clicks save and continue' do
```
wmitchell-cambria: Another reason to investigate how we are implementing this... These regression tests are unsustainable. Regression/acceptance tests are meant to cover the happy paths. They take a long time to set up and a long time to run. Having three scenarios for this _link_ (It's just a link!) is ridiculous. Multiply that by breadcrumbs, print button, etc... we will never be able to run them all consistently.

For now, I would recommend combining these into a single scenario, or covering the cases more thoroughly in unit tests.

app/javascript/Application/components/Assessment/style.sass
wmitchell-cambria: I just looked at the design, and I am still not convinced. It 100% looks like it should be positioned on the left. Have you looked into using Row and Col from component library to position the buttons?

wmitchell-cambria: The reason I am focusing on this is two-fold. First, it will break on anything other than the one screen size you tested with. Second, we are currently in-flight on a story that changes rem, which will also break this style even if the screen resolution stays the same. Positioning from the left will be much more robust.

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
nasar-cambria: Yes

spec/regression/case_worker_scenarios_spec.rb
nasar-cambria: removed.

spec/regression/case_worker_scenarios_spec.rb
nasar-cambria: This was asked as per the suggestion from Stan (QA) and mentioned in the test script as well.  I can combine it however I need to talk with him and the team to change about your suggestions and then I can come back and make the change.

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -2,14 +2,16 @@ import React, { PureComponent } from 'react'
...
 
 import './style.sass'
 
 class AssessmentFormFooter extends PureComponent {
   render() {
-    const { isSubmitButtonEnabled, onCancelClick, onSubmitAssessment } = this.props
+    const { assessment, isSubmitButtonEnabled, onCancelClick, onSubmitAssessment, onClick } = this.props
     return (
       <div className={'form-footer'}>
+        <AssessmentChangelogLink onClick={onClick} assessmentId={assessment.id} assessmentStatus={assessment.status} />
```
AK56: I believe we can remove the `onClick` prop here, since neither it seem to be passed in from the parent component, which in this case is `AssessmentContainerInner` nor it is being utilized by the child component `AssessmentChangelogLink`

app/javascript/Application/components/Assessment/style.sass
nasar-cambria: I tried using Column ```Col``` and ```Row``` from the component library however it is not setting ```Complete``` and ```Cancel``` button accurately in the same position that they are in right now. I can show a demo in a separate pr if it makes sense we can merge it in or discard the changes.

Also, I remember now during refinement we discussed that this link is here temporarily until we get a new design for it.

PR #545, opened 2019-02-25
CANS-949 Show Change Log Link for Client from another county

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-949

## Description
Implantation for New AC to show View Change Log link for client from another county and remove Discard continue button from the modal if the Assessment is new (Assessment that has never been saved)

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -368,6 +368,7 @@ export default class AssessmentContainer extends Component {
             discardAndContinue={this.loadAssessment}
             saveAndContinue={this.handleSaveAssessment}
             isUnsaved={this.state.isUnsaved}
+            assessmentId={this.props.match.params.id}
```
wmitchell-cambria: The prop types for `match` do not guarantee that these params will exist. It only defines it as an object. We should either pass the match through entirely, or define the prop types for what the component expects.

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,35 +1,63 @@
...
 
 import './style.sass'
 
 class AssessmentFormFooter extends PureComponent {
+  displayFooterButtons() {
+    const { assessment, isEditable, onCancelClick, isSubmitButtonEnabled, onSubmitAssessment } = this.props
+    const isUnderSix = assessment && assessment.state && assessment.state.under_six
+
+    return (
+      <Fragment>
```
wmitchell-cambria: Fragment is unnecessary here. Instead, we can just do
```javascript
return isUnderSix !== undefined && isEditable ? (<div>...</div>) : null
```

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,35 +1,63 @@
...
-      <div className={'form-footer'}>
-        <AssessmentChangelogLink assessmentId={assessment.id} assessmentStatus={assessment.status} />
-        <Button id="cancel-assessment" color="secondary" className="button-fix-secondary" onClick={onCancelClick}>
-          Cancel
-        </Button>
-        <CompleteAssessmentButton onSubmitAssessment={onSubmitAssessment} disabled={!isSubmitButtonEnabled} />
-      </div>
+      <Row className={'form-footer'}>
+        <Col sm={6}>
+          {isUnderSix !== undefined ? (
```
wmitchell-cambria: It's not very obvious why `isUnderSix` is being checked here. We also do this check in two different places. It may be cleaner if we named it something different like `isFooterVisible` and passed it in directly as a prop.

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,35 +1,63 @@
...
+            <AssessmentChangelogLink assessmentId={assessment.id} assessmentStatus={assessment.status} />
+          ) : null}
+        </Col>
+        <Col sm={6}>{this.displayFooterButtons()}</Col>
```
wmitchell-cambria: Just because you factored this out into a helper method, I am obliged to comment... If it is complex enough for a helper method, it is probably complex enough to have its own component. Small components should be our primary form of factoring out code. Note that here we also created a separate `describe` block in the unit tests. So, this method that returns JSX already has its own unit tests, too!

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
```diff
@@ -1,35 +1,63 @@
...
+  isEditable: PropTypes.bool,
+  isSubmitButtonEnabled: PropTypes.bool,
   onCancelClick: PropTypes.func.isRequired,
   onSubmitAssessment: PropTypes.func.isRequired,
 }
 
 AssessmentFormFooter.defaultProps = {
   assessment: {},
+  isSubmitButtonEnabled: true,
+  isEditable: undefined,
```
wmitchell-cambria: Possibly `false` would be more obvious? Also, are we sure that we want the submit button to default to enabled? I haven't thought through all of the states, but it sounds like that could allow the user to submit something before it is ready (perhaps if some data is not fully loaded yet).

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -18,9 +18,12 @@ $item-confidential-label-height: 6.05rem
   margin-left: 0.5rem
 
 .view-changelog-link
-  margin-right: 49.375rem
   text-transform: uppercase
 
+.rightButton
```
wmitchell-cambria: CSS class names should be hyphenated, not camel case. They are not typically case sensitive. Also, this class name itself is a bit misleading, as it applies to a `div`, not a `button`.

app/javascript/Application/components/common/UnsavedDataWarning.js
```diff
@@ -34,6 +35,11 @@ class UnsavedDataWarning extends Component {
     } else {
       eventBus.post(event)
     }
+    if (this.props.assessmentId) {
```
wmitchell-cambria: The existence of the assessmentId is not an accurate reflection of whether the assessment is rendered. It tells whether the assessment has been saved. In the case of an unsaved new assessment, there could be rendered changes on the page, but no assessmentId yet. The naming of the `isAssessmentRendered` state property is confusing.

app/javascript/Application/components/common/UnsavedDataWarning.js
```diff
@@ -72,27 +78,32 @@ class UnsavedDataWarning extends Component {
...
-          <Button
-            className={'unsaved-warning-modal-discard'}
-            onClick={() => {
-              this.onAction(this.props.discardAndContinue)
-            }}
-          >
-            {'Discard changes and continue'}
-          </Button>
-        </ModalFooter>
+        {this.state.isAssessmentRendered ? (
```
wmitchell-cambria: I am surprised that there are not more unit test changes for this file.

nasar-cambria: Added new test case to validate ```isAssessmentSaved```

app/javascript/Application/components/Assessment/AssessmentFormFooter.js
nasar-cambria: We need this check to render footer link ```view change log``` after the age button is selected.  I think keeping it isUnderSix would be better to understand and consistent because we are using the same check in different places.

nasar-cambria:
@wmitchell-cambria ```view-change-log``` link has a save button inside the modal and a user can save the form when he clicks on ```view-change-log``` for the first time. The idea is to not hide this link even on the new assessment and if the user wants to navigate he would have to click on ```Save and Continue``` in order to redirect to change log.

app/javascript/Application/components/common/UnsavedDataWarning.js
nasar-cambria: changed name to ```isAssessmentSaved```

nasar-cambria:
Jenkins re-test this please

nasar-cambria:
@wmitchell-cambria when you get a chance can you please revisit?

wmitchell-cambria:
Catching up now.

app/javascript/Application/components/common/UnsavedDataWarning.test.js
```diff
@@ -71,5 +84,10 @@ describe('<UnsavedDataWarning />', () => {
       expect(wrapper.instance().props.discardAndContinue).toBeCalled()
       expect(wrapper.instance().state.isOpened).toBeFalsy()
     })
+
+    it('"should not render discardAndContinue button on a unsaved assessment', () => {
+      const wrapper = unsavedAssessment()
+      expect(wrapper.instance().state.isAssessmentSaved).toBeFalsy()
```
wmitchell-cambria: nits:  `wrapper.instance().state` is the same as the shorter `wrapper.state()`. Also, small typo in the test name with the `"`

PR #550, opened 2019-02-27
CANS-1005 Assessment comparison

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1005

## Description

- This story will be no additional regression test kit required, confirmed with QA already.
- Its functionalities based on the JIRA Ticket acceptance criteria, except it is "NOT" a new (separate) page which confirmed with design team already.
## Tests

## Types of changes

## Checklist:


zliuwd:
retest this please

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+  )
+}
+
+AssessmentComparison.propTypes = {
+  comparisonRecords: PropTypes.array,
+  recordsModeSwitch: PropTypes.func.isRequired,
+}
+
+AssessmentComparison.defaultProps = {
+  comparisonRecords: undefined,
```
denys-davydov: Since in `#render()` you have `props.comparisonRecords.length` and the prop is not required, please consider to have an empty array `[]` as a default value for the `comparisonRecords` prop.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+import RecordsModeSwitchButton from '../RecordsModeSwitchButton'
+import { recordsMode } from '../Client.helper'
+
+const AssessmentComparison = props => {
+  if (props.comparisonRecords.length < 1) {
+    return null
+  }
+  const data = props.comparisonRecords[0]
+  return (
+    <Fragment>
```
denys-davydov: There is no need to have a `<Fragment/>` with 1 child. Use it when you have multiple children and do not want to wrap them with a `<div/>`. Just omit the Fragment here.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+      <Grid item xs={12}>
```
denys-davydov: Not sure if you need a `<Grid/>` here as well. Try to omit it and see the difference.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+            <CardTitle>
+              <span>Assessment Comparison</span>
+              <RecordsModeSwitchButton
+                switchButtonName={recordsMode.HISTORY}
+                recordsModeSwitch={props.recordsModeSwitch}
+              />
+            </CardTitle>
+          </CardHeader>
+          <CardBody className={'comparison-card-body'}>
+            <ComparisonOuterTable data={data} i18n={props.comparisonRecords[1]} />
```
denys-davydov: So `props.comparisonRecords` is an array of objects with different types: element 0 is some `data`, while element 1 is an i18n object. It looks like you don't need an array for this situation, consider to simply have an object with fields or (which is preferable) to have 2 different props for element 0 and element 1 with properly described propTypes and defaultProps.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import LoadingBoundary from '../../common/LoadingBoundary'
+import { I18nService } from '../../Assessment/'
+import { ClientService } from '../'
+
+class AssessmentComparisonLoadingBoundary extends React.Component {
+  state = {}
+
+  static getDerivedStateFromProps({ clientIdentifier: propsId, instrumentId }, { id: stateId }) {
```
denys-davydov: Usage of `getDerivedStateFromProps()` looks pretty complicated. Is there a way to omit it? And to handle clientIdentifier presence in `render()` function?

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
```diff
@@ -0,0 +1,115 @@
...
+  getTitle,
+  expander,
+  domainTitleSwitcher,
+  blankColFiller,
+  requiredDomainColumnsAmount,
+  domainRatingSwitcher,
+} from './comparisonHelper'
+import './style.sass'
+
+class ComparisonOuterTable extends React.Component {
```
wmitchell-cambria: How is the performance of rendering these? Since they are looping over a lot of data, there may be some lag while React processes the whole render tree. Since these are displaying the data without any side effects, this could be a situation where using `React.PureComponent` has a big impact. I'd consider using it for the InnerTable as well. It will allow React to skip re-rendering all of these tables when the data doesn't change.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
```diff
@@ -0,0 +1,115 @@
...
+      Header: 'Domain Name',
+      sortable: true,
+      accessor: domain => {
+        const domainTitle = getTitle(this.props.i18n, domain.code)
+        const finalTitle = domainTitleSwitcher(domain, domainTitle)
+        return finalTitle
+      },
+      style: {
+        height: '3rem',
+        ...verticalCenter,
```
wmitchell-cambria: Same here. Can we keep styles in CSS?

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.test.js
```diff
@@ -0,0 +1,100 @@
...
+    const target = wrapper.find(DataGrid).props().columns[1].Header
+    expect(mount(target).text()).toBe(expectedDate)
+  })
+
+  it('will render ComparisonOuterTableExpander ', () => {
+    const target = wrapper.find('.outer-table-expander-collapsed')
+    expect(target.exists()).toBe(true)
+  })
+
+  it(' ComparisonInnerTable will show after ComparisonOuterTableExpander be clicked', () => {
```
wmitchell-cambria: nit: Extra space at the start of many of these test descriptions. Also, 'is clicked' instead of 'be clicked'.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableExpander.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Icon from '@cwds/icons'
+
+const ComparisonOuterTableExpander = ({ isExpanded }) => {
+  return (
+    <div>
+      {isExpanded ? (
+        <span>
```
wmitchell-cambria: Are both the `div` and `span` necessary? Seems like extra HTML to me. (Extra HTML, repeated often, can really slow down a page.)

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableHeader.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { isoToLocalDate } from '../../../util/dateHelper'
+
+const ComparisonOuterTableHeader = ({ index, dateInfo }) => {
```
wmitchell-cambria: If possible, I think we should just pass in the single event_date that we need here, instead of the entire list with an index. Passing only the data required for a component is good practice and helps to ensure it doesn't render more often than necessary.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableHeader.test.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import ComparisonOuterTableHeader from './ComparisonOuterTableHeader'
+import { isoToLocalDate } from '../../../util/dateHelper'
+
+describe('<ComparisonOuterTableHeader />', () => {
+  const inProgressProps = {
+    index: 0,
```
wmitchell-cambria: If we do keep it with separate index and array props, we should at least test that they work by using something other than 0 in some cases.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
+import ComparisonOuterTableExpander from './ComparisonOuterTableExpander'
+import { getI18nByCode } from '../../Assessment/I18nHelper'
```
wmitchell-cambria: Another reason to move i18n code to a common directory. This reaches across domain boundaries by going _into_ the Assessment folder.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+        result = `Caregiver Domain-${domain.caregiver_index}`
+      }
+      break
+    default:
+      result = domainTitle
+  }
+  return result
+}
+
+export const domainRatingSwitcher = (ratingTotal, code) => {
```
wmitchell-cambria: nit: `domainRatingSwitcher` does not say much about what this function does. A reader of the component has to look at this file to see the code. Naming this is also difficult, because it is doing two different things. It is replacing empty values with `--`, and it is adding `Y` to positive boolean ratings. Maybe we can separate them into two functions, so that the calling code is more like `const displayRating = addYesToBooleanCodes(mapNullsToDashes(totalRating), code)`

Maybe that isn't the exactly right solution... but I do think some of these helpers are named very abstractly. Having better names would make the code overall easier to understand.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+    return '--'
+  } else if (code === 'TRM') {
+    return `${ratingTotal}-Y`
+  } else {
+    return ratingTotal
+  }
+}
+
+const itemRatingGenerator = (itemRatings, index) => {
+  const naRating = 8
```
wmitchell-cambria: This constant should only be set in one location in our code base. We can import it from there (and move it to a shared location if needed).

app/javascript/Application/components/Client/AssessmentComparison/style.sass
```diff
@@ -0,0 +1,43 @@
...
+.outer-content
+  text-align: center
+  font-size: 1.1rem
+
+.outer-content-header
+  text-align: center
+  font-size: 1.1rem
+  margin-bottom: 0.7rem
+  padding-top: 0rem !important
+  display: 'flex'
```
wmitchell-cambria: I don't think these need to be in quotes.

app/javascript/Application/components/Client/Client.helper.test.js
```diff
@@ -193,3 +198,24 @@ describe('All mock data are defined', () => {
     expect(childInfoJson).toBeDefined()
   })
 })
+
+describe('shouldRenderRecordsSwitch', () => {
+  const towOrMoreValidAssessments = [
```
wmitchell-cambria: nit: tow => two

app/javascript/Application/components/Client/Client.service.js
```diff
@@ -4,7 +4,9 @@ export class ClientService {
   static fetch(id) {
     return apiGet(`/clients/${id}`)
   }
-
+  static getComparison(id) {
```
wmitchell-cambria: nit: `getAssessmentComparison` would match the API call a little better.

app/javascript/Application/components/Client/RecordsModeSwitchButton.js
```diff
@@ -0,0 +1,33 @@
+import { recordsMode, shouldRenderRecordsSwitch } from './Client.helper'
+import { Button } from '@cwds/components'
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const RecordsModeSwitchButton = ({ switchButtonName, recordsModeSwitch, assessments }) => {
+  const switchButton = (
+    <span>
```
wmitchell-cambria: Is the span wrapper necessary?

app/javascript/Application/components/Client/RecordsModeSwitchButton.test.js
```diff
@@ -0,0 +1,86 @@
...
+describe('</RecordsModeSwitchButton>', () => {
+  afterEach(() => {
+    wrapper.unmount()
+  })
+
+  it('will render a button with correct text, when there are two or more assessments which status are not DELETED ', () => {
+    getWrapper(defaultFakeProps)
+    const target = wrapper.find(Button)
+    expect(target.length).toBe(1)
+    expect(target.text()).toBe('recordsMode.COMPARISION')
```
wmitchell-cambria: typo: COMPARISION => COMPARISON

app/javascript/Application/components/Client/style.sass
```diff
@@ -55,3 +55,6 @@
     -ms-overflow-style: -ms-autohiding-scrollbar
   #no-data
     margin-left: 3.5rem
+.records-mode-switch-button
```
wmitchell-cambria: Why do these buttons have custom font-sizes? Shouldn't they inherit their styles from the component library?

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
wmitchell-cambria: I agree that it's complicated, but I'm not sure there is a way around it. It's the standard way in React to watch for changes in props, and that's necessary for LoadingBoundaries so that they know when to refetch their data (without refetching it every time). I think this is an OK use case for it.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
```diff
@@ -0,0 +1,38 @@
...
+    if (propsId !== stateId) {
+      return {
+        id: propsId,
+        fetch: () => Promise.all([ClientService.getComparison(propsId), I18nService.fetchByInstrumentId(instrumentId)]),
```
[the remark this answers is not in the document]
wmitchell-cambria: Also, this is where you would want to map from an array to an object with keys. Something like this:

```javascript
fetch: () =>
  Promise.all([ClientService.getComparison(propsId), I18nService.fetchByInstrumentId(instrumentId)])
    .then(([data, i18n]) => ({ data, i18n })),
```

zliuwd:
Thanks for the reviews and comments from everyone. I will resolve these comments in tomorrow.
 Down below is the test coverage report:
screen shot 2019-02-27 at 4 36 29 pm

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+      result = domainTitle
+  }
+  return result
+}
+
+export const domainRatingSwitcher = (ratingTotal, code) => {
+  if (!ratingTotal) {
+    return '--'
+  } else if (code === 'TRM') {
+    return `${ratingTotal}-Y`
```
LeonidMarushevskyi: API returns attribute 'type' for both domain and rating with possible values 'BOOLEAN', 'REGULAR'. Better to use it to decide the representation instead of hardcoding domain code here.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+  const checker = amount - targetCols.length
+  if (checker > 0) {
+    Array.from({ length: checker }).forEach(() => {
+      targetCols.push(blankColumn(keyWords))
+    })
+  }
+}
+
+export const domainTitleSwitcher = (domain, domainTitle) => {
```
LeonidMarushevskyi: I don't think we need to have this domain name overriding here at all. We can just keep domain names as in regular form. With future changes related to age toggle on this page, there will be no need to differentiate domains with identical titles but different codes. So no need to add age group to the domain name. 

Also no need to hardcode caregiver domain code. You can just add caregiver names for domains with is_caregiver_domain = true.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+const itemRatingGenerator = (itemRatings, index) => {
+  const naRating = 8
+  let rating = ''
+  if (
+    itemRatings[index] &&
+    itemRatings[index].value !== undefined &&
+    itemRatings[index].value !== -1 &&
+    itemRatings[index].value !== naRating
```
LeonidMarushevskyi: You don't need to do validation for -1 and 8 here. You can completely rely on API to provide the correct rating. API will not return 'value' attribute if not rating available for domain or item.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
zliuwd: use Container from design lib instead.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
zliuwd: Hi Deny, thanks for the comment about this point, just checked our code base, many 
Boundary components use this, adding on Will's comment, I will keep it as is, of course, I would like to refactor it, if we have any better approaching in the future.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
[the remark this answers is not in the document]
zliuwd: Done, replace it with Container from lib.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+import Grid from '@material-ui/core/Grid'
+import { Card, CardHeader, CardBody, CardTitle } from '@cwds/components'
+import ComparisonOuterTable from './ComparisonOuterTable'
+import './style.sass'
+import PropTypes from 'prop-types'
+import RecordsModeSwitchButton from '../RecordsModeSwitchButton'
+import { recordsMode } from '../Client.helper'
+
+const AssessmentComparison = props => {
+  if (props.comparisonRecords.length < 1) {
```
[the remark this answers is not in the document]
zliuwd: Done, replaced with [ ]

app/javascript/Application/components/Client/Client.helper.test.js
zliuwd: done

app/javascript/Application/components/Client/RecordsModeSwitchButton.js
zliuwd: Removed

app/javascript/Application/components/Client/RecordsModeSwitchButton.test.js
zliuwd: Good catch, thanks!

app/javascript/Application/components/Client/style.sass
zliuwd: 😜 I just like the small font button, I removed it now.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
```diff
@@ -0,0 +1,115 @@
...
+    this.state = {}
+  }
+
+  domainTotalColsGenerator = counter => {
+    return counter.map((el, index) => {
+      return {
+        id: `domain-assessment-${index}`,
+        style: {
+          minWidth: '8rem',
+          height: '3rem',
```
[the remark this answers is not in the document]
zliuwd: minWidth could be removed, but height have to be here. Style prop is the only way I could find to adjust the row height. But I believe this will be removed in next sprint after we removed all the "In progress" assessments data. At that time, we do not need to display status in the header then we don't need adjust the row height.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,83 @@
...
+    super(props)
+    this.state = {}
+  }
+
+  itemRatingColsGenerator = () => {
+    return this.props.counter.map((el, index) => {
+      return {
+        id: `item-assessment-${index}`,
+        style: {
+          minWidth: '8rem',
```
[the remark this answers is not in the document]
zliuwd: removed

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.test.js
zliuwd: 👍

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,46 @@
...
+                recordsModeSwitch={props.recordsModeSwitch}
+              />
+            </CardTitle>
+          </CardHeader>
+          <CardBody className={'comparison-card-body'}>
+            <ComparisonOuterTable data={data} i18n={props.comparisonRecords[1]} />
+          </CardBody>
+        </Card>
+      </Grid>
+    </Fragment>
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/style.sass
zliuwd: 👍

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableExpander.js
zliuwd: removed

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
```diff
@@ -0,0 +1,114 @@
...
+
+  domainTotalColsGenerator = counter => {
+    return counter.map((el, index) => {
+      return {
+        id: `domain-assessment-${index}`,
+        style: {
+          height: '3rem',
+          ...verticalCenter,
+        },
+        Header: <ComparisonOuterTableHeader index={index} dateInfo={this.props.data.event_dates} />,
```
denys-davydov: If my understand is correct, we send the whole array `this.props.data.event_dates` and the element index to the `ComparisonOuterTableHeader` component and then inside of the component we get the element from the array by the index. It looks redundant since we already have the array element in the `el` variable, so we can simply send the element to the `ComparisonOuterTableHeader` component.

zliuwd:
### For the question "is Client the right directory for it":

I still believe Client is the right directory for it, based on some points down below:
Looks we have two rules for that, first is layout structures , second is the functionalities. Let we talk about functionalities. 

All CANS are focus on the assessment functionalities, so when we have the directory Assessment and other directories like Common, Client, Print, Search and Staff at the same time, we should identify what is the "Prerequisites" for the directory, because all of them are dealing with assessments , otherwise we should put all the stuff in Assessment directory. In my opinion, there are couple "Prerequisites" as below:

###  1. Primary and secondary functionalities about the assessment (core and assistance)
Based on current directories. We could see all the functionalities about create, delete and edit assessments are locating in Assessment directory, but the functionalities about search are distributed in other directories, the reason, I think it is search will not change the data, They are considered secondary or assistance functionalities. Assessment Comparison actually does not really change anything also, It should not in the Assessment directory, an it should existe like how History, Search, Staff, Print existed in our App.

### 2. “Access” is another obvious prerequisite. 
For example, “user could see print mode if and only if he click the print button”, “user could see client’s history if and only if he access client profile ”, I know from Search we also can access client history, but we still need get client profile page firstly. Currently, “Assessment comparison” could be shown only when a user access client profile, in other words, “Assessment comparison” could only be accessed by client id, it is kind of private functionality of Client.

### 3. Functional Independence
“Assessment Comparison” with very high Independence, it has nothing imported from Assessment (after moved i18n to common ) and with individual API, but it wrapped up by Client.js. This also a reason it should be in Client directory.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,80 @@
...
+    const items = this.props.items
+    const TheadComponent = props => null
+
+    return (
+      <DataGrid
+        key={this.props.domainCode}
+        columns={itemCmparisionTableTemplate}
+        data={items}
+        sortable={false}
+        minRows={gridMinRows(this.props.items)}
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,80 @@
...
+        return itemTitle
+      },
+      className: 'inner-item-name-content',
+    }
+
+    const itemRatingCols = this.itemRatingColsGenerator()
+    blankColFiller(itemRatingCols, 'item', requiredDomainColumnsAmount)
+    const itemCmparisionTableTemplate = [itemNameCol, ...itemRatingCols, expanderPlaceHolder]
+    const items = this.props.items
+    const TheadComponent = props => null
```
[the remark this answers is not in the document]
zliuwd: done

wmitchell-cambria:
I think that reasoning makes sense. My biggest reasons for having it in Assessment were i18n and domain code dependencies. If those are no longer an issue, then Client seems like a reasonable home to me.

app/javascript/Application/components/Client/AssessmentComparison/style.sass
```diff
@@ -0,0 +1,43 @@
...
+.outer-content
+  text-align: center
+  font-size: 1.1rem
+
+.outer-content-header
+  text-align: center
+  font-size: 1.1rem
+  margin-bottom: 0.7rem
+  padding-top: 0rem !important
+  flexDirection: column
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/style.sass
```diff
@@ -0,0 +1,43 @@
...
+  justifyContent: center
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export const expanderPlaceHolder = {
+  id: 'expander-place-holder',
+  width: CP_TABLE_COL_WIDTHS.EXPANDER,
+}
+
+export const requiredDomainColumnsAmount = 5
+
+export const blankColFiller = (targetCols, keyWords, amount) => {
```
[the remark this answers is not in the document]
zliuwd: 👍

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+  return null
+}
+
+export const expander = {
+  id: 'expander',
+  width: CP_TABLE_COL_WIDTHS.EXPANDER,
+  expander: true,
+  Expander: ComparisonOuterTableExpander,
+  style: {
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableHeader.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: About domain name, we need keep as-is and refactor in next sprint because the AC of this story. I will change to 
is_caregiver_domain for second concern.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -0,0 +1,43 @@
...
+        <CardBody className={'comparison-card-body'}>
+          <ComparisonOuterTable data={data} i18n={props.comparisonRecords[1]} />
+        </CardBody>
+      </Card>
+    </Container>
+  )
+}
+
+AssessmentComparison.propTypes = {
+  comparisonRecords: PropTypes.array,
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,80 @@
...
+      accessor: item => {
+        const itemTitle = getTitle(this.props.i18n, item.code)
+        return itemTitle
+      },
+      className: 'inner-item-name-content',
+    }
+
+    const itemRatingCols = this.itemRatingColsGenerator()
+    blankColFiller(itemRatingCols, 'item', requiredDomainColumnsAmount)
+    const itemCmparisionTableTemplate = [itemNameCol, ...itemRatingCols, expanderPlaceHolder]
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,80 @@
...
+    const items = this.props.items
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
```diff
@@ -0,0 +1,114 @@
...
+        i18n={this.props.i18n}
+      />
+    )
+  }
+
+  render() {
+    if (!this.props.data) {
+      return null
+    }
+    const counter = this.props.data.event_dates
```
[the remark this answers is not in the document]
zliuwd: changed the name.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
zliuwd: Yeah, changed

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,80 @@
...
+        headerClassName: 'text-center',
+      }
+    })
+  }
+
+  render() {
+    const itemNameCol = {
+      id: 'itemName',
+      width: CP_TABLE_COL_WIDTHS.ITEM_NAME,
+      accessor: item => {
```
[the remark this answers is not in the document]
zliuwd: 👍

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.js
```diff
@@ -0,0 +1,83 @@
...
+      />
+    )
+  }
+}
+
+ComparisonInnerTable.propTypes = {
+  counter: PropTypes.array.isRequired,
+  domainCode: PropTypes.string.isRequired,
+  i18n: PropTypes.object,
+  items: PropTypes.array.isRequired,
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.test.js
```diff
@@ -0,0 +1,57 @@
...
+import * as helper from './comparisonHelper'
+const fakeProps = {
+  key: 'somekey',
+  domainCode: 'TRM',
+  items: fakeData.domains[0].items,
+  counter: fakeData.event_dates,
+  i18n: { key: 'value' },
+}
+
+describe('<ComparisonInnerTable />', () => {
```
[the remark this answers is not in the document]
zliuwd: I already change the "counter" to "assessmentDates", the using is: generating the cols according to valid assessments. For example, now our cols requirement for this form is 5, if we just have 3 valid assessments, we will need two other blank cols, so we use the counter to map the 3 valid cols then use blankColFiller to add on another two blank cols, then we get the template. we don't want the dynamic adjusting width of cols, so we go this way.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonInnerTable.test.js
```diff
@@ -0,0 +1,57 @@
...
+  afterEach(() => {
+    wrapper.unmount()
+  })
+
+  it('will render a DataGrid', () => {
+    const target = wrapper.find(DataGrid)
+    expect(target.length).toBe(1)
+  })
+
+  it('will render a DataGrid with 7 columns', () => {
```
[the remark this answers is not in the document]
zliuwd: Good idea, should be always 7, this test is just for making sure this.

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTable.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: Separated, and named "handleCaregiverDomainTitle", by the way this case switch will be totally removed in next sprint, because we decided use "full title" int todays refinment

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,118 @@
...
+    Array.from({ length: checker }).forEach(() => {
+      targetCols.push(blankColumn(keyWords))
+    })
+  }
+}
+
+export const domainTitleSwitcher = (domain, domainTitle) => {
+  let result
+  switch (domain.code) {
+    case 'TRM':
```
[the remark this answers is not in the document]
zliuwd: For this form these code just temporarily use and will removed next sprint, I just refactor all of them include which AssessmentSummary used to util folder constants.js.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: removed

app/javascript/Application/components/Client/Client.service.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/ComparisonOuterTableHeader.test.js
zliuwd: this component already be refactored as accept status and date, we don't need index, then no need test it.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.test.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import AssessmentComparisonLoadingBoundary from './AssessmentComparisonLoadingBoundary'
+import LoadingBoundary from '../../common/LoadingBoundary'
+import { I18nService } from '../../Assessment/'
```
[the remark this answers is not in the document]
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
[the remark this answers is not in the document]
wmitchell-cambria: Refactoring to get rid of them seems good. While we are using them, though, can we please import them from the domain helper file? That way if the Tech Debt story is deprioritized or left unfinished, we will at least have these consolidated into a single file.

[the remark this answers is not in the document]
zliuwd: Got it, moved.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
```diff
@@ -0,0 +1,114 @@
...
+
+export const blankColFiller = (targetCols, keyWords, amount) => {
+  const blankColumn = { id: `${keyWords}-blank-column`, header: null, sortable: false }
+  const checker = amount - targetCols.length
+  if (checker > 0) [...Array(checker)].map(() => targetCols.push(blankColumn))
+}
+
+const dashes = '--'
+
+const handleCaregiverDomainTitle = (domain, domainTitle) => {
```
denys-davydov: `domainTitle` parameter is not used in code, please remove it

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import LoadingBoundary from '../../common/LoadingBoundary'
+import { I18nService } from '../../common/'
+import { ClientService } from '../'
+
+const AssessmentComparisonLoadingBoundary = ({ children, clientIdentifier, instrumentId }) => {
+  const fetch = () =>
```
wmitchell-cambria: When we define the `fetch` function inside a render method like this, the LoadingBoundary will believe that fetch is changing every time it renders, causing it to refetch the data. _Especially_ in a functional component (with a functional parent), this can lead to a lot of excess network requests.

That is why we use a `getDerivedStateFromProps` lifecycle method in these components that depend on external loading. We set the fetch function once for a set of props, and we ensure that the function reference only changes when the props themselves change, not each render. It's one of the [documented exceptions](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change) to "You Probably Don't Need Derived State".

This component was written right the first time around.

app/javascript/Application/components/Assessment/AssessmentSummary/DomainHelper.js
```diff
@@ -12,9 +12,16 @@ export const DomainPropType = PropTypes.shape({
 
 export const DomainsPropType = PropTypes.arrayOf(DomainPropType)
 
-const STRENGTH_DOMAINS = ['STR', 'EST']
-const TRAUMA_DOMAIN = 'TRM'
-const BEN_DOMAIN = 'BEN'
+export const domainCodes = Object.freeze({
+  STRENGTH_DOMAIN_UNDER_SIX: 'EST',
+  STRENGTH_DOMAIN_UPON_SIX: 'STR',
```
wmitchell-cambria: nit: I think most fields we use for this say 'above' instead of 'upon'.

app/javascript/Application/components/Client/AssessmentComparison/comparisonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
zliuwd: Reversed it, and very nice to know the difference. :)

app/javascript/Application/components/Assessment/AssessmentSummary/DomainHelper.js
zliuwd: changed

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types'
+import LoadingBoundary from '../../common/LoadingBoundary'
+import { I18nService } from '../../common/'
+import { ClientService } from '../'
+
+class AssessmentComparisonLoadingBoundary extends React.Component {
+  state = {}
+
+  static getDerivedStateFromProps({ clientIdentifier: propsId, instrumentId }, { id: stateId }) {
+    if (propsId !== stateId) {
```
wmitchell-cambria: One other comment I can add here is that it may be possible to improve the readability of these functions. At first I liked the `propsId/stateId` pattern, but it may end up making it appear more complicated. Instead, the recent loading boundaries I have written, I have avoided the destructuring. That, combined with moving the fetch to a helper factory method, can make it more obvious what is going on:

```javascript
// Near the top of the file:
const fetchFactory = (clientId, instrumentId) => () =>
  Promise.all([
            ClientService.getAssessmentComparison(props.clientIdentifier),
            I18nService.fetchByInstrumentId(props.instrumentId),
          ]).then(([data, i18n]) => ({ data, i18n }))

// ...

static getDerivedStateFromProps(props, state) {
    if (props.clientIdentifier !== state.id) {
      return {
        id: props.clientIdentifier,
        fetch: fetchFactory(props.clientIdentifier, props.instrumentId),
      }
    }
    return null
}
```

Anyways, they are functionally the same thing. Just a matter of style preference, so your call.

zliuwd:
retest this please

zliuwd:
retest this please

PR #552, opened 2019-02-28
CANS-1070: Use React "History" approach for unsaved assessment warning

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


DmitriyRudenko:
Jenkins retest this please

app/javascript/Application/components/common/ExternalLink.js
```diff
@@ -0,0 +1,36 @@
+import pageLockService from './PageLockService'
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+class ExternalLink extends Component {
+  constructor(context) {
+    super(context)
+    this.confirmLeavePage = this.confirmLeavePage.bind(this)
```
wmitchell-cambria: One pattern that we can start using more is to use arrow functions instead of methods, to avoid the need to do this binding. Example:

```javascript
confirmLeavePage = event => {
  event.preventDefault()
  pageLockService.confirm(this.leavePage) // `this` is already bound.
}
```

In this case, it would let us get rid of the constructor entirely.

app/javascript/Application/components/common/PageLockService.js
```diff
@@ -0,0 +1,67 @@
+import { createBrowserHistory } from 'history'
+import { basePath } from '../../util/common'
+
+class PageLockService {
+  // creates react history, sets page unloading confirmation
+  // confirmation will be triggered ONLY IF method lock() has been called before
+  constructor() {
```
wmitchell-cambria: I'm pretty sure all constructors should call `super()`. If this is a singleton, though, I would avoid using `class` and just define it as a plain object literal.

app/javascript/Application/components/common/ExternalLink.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  leavePage() {
+    pageLockService.unlock()
+    document.location.assign(this.props.href)
+  }
+
+  render() {
+    return (
+      <a href={this.props.href} onClick={this.confirmLeavePage}>
```
wmitchell-cambria: I'm not so sure about this ExternalLink. Overriding normal link behavior, just to get a custom onbeforeunload modal seems extreme. One of the things I liked about this approach is that regular links would fall back to the default browser behavior. Can we push back on Product with this one? Fighting against browser behavior like this is a losing battle.

app/javascript/Application/components/common/PageLockService.test.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(pageLockService.pageLock.tryAction).toBe(tryAction)
+    expect(pageLockService.pageLock.unblock).toBe(unblock)
+  })
+
+  it('verify unlock', () => {
+    pageLockService.unlock()
+    expect(pageLockService.pageLock).toBeUndefined()
+  })
+
+  it('verify locked confirm', () => {
```
wmitchell-cambria: Test names should read like sentences and describe what the expected behavior is. For example, `it('tries the action when locked and confirmed')`. This helps future readers understand what behaviors are required, which is important for refactoring.

app/javascript/Application/components/common/UnsavedDataWarning.test.js
```diff
@@ -51,7 +50,26 @@ describe('<UnsavedDataWarning />', () => {
...
+    it('page is unlocked when assessment is saved', () => {
+      pageLockService.unlock()
+      openWarning(false)
+      expect(pageLockService.pageLock).toBeUndefined()
+    })
+
+    it('page is locked when assessment is unsaved', () => {
+      pageLockService.unlock()
+      openWarning(true)
+      expect(pageLockService.pageLock).not.toBeUndefined()
```
wmitchell-cambria: nit: `.toBeDefined()`

wmitchell-cambria:
Chatted with @DmitriyRudenko offline to talk through the comments. He's going to address most of them, so I am dropping my Request Changes in case I am not around to review again later.

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
```diff
@@ -1,30 +1,16 @@
...
-  }
-
-  canRedirectAssessment = () => {
-    this.setState({ shouldRedirect: true })
-  }
-
-  onClick = event => {
-    event.preventDefault()
-    eventBus.post(UNSAVED_ASSESSMENT_VALIDATION_EVENT, CHANGE_LOG_EVENT)
+  onClick = () => {
```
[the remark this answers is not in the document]
DmitriyRudenko: good catch, this component is removed

app/javascript/Application/components/common/PageLockService.js
DmitriyRudenko: it doesn't extend anything

app/javascript/Application/components/common/ExternalLink.js
DmitriyRudenko: this is invented just to not duplicate external link logic

app/javascript/Application/components/common/PageLockService.test.js
DmitriyRudenko: done

app/javascript/Application/components/common/UnsavedDataWarning.test.js
DmitriyRudenko: done

app/javascript/Application/components/Assessment/AssessmentChangelogLink.js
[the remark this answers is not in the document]
DmitriyRudenko: reverted it back:-( Unfortunately there are problems with "mount" AssessmentContainer in unit tests. It needs to be researched.

PR #556, opened 2019-03-02
CANS-1052-Implemented Sorting for DOB in client list card

KShresthaCwds:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1052
## Description
Implemented Sorting for DOB in client list card
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
```diff
@@ -25,7 +25,7 @@ and meet the setting rules of reactTable at same time */
       },
       accessor: client => renderDate(client.dob),
       className: 'client-list-table-cell-center',
-      sortable: true,
+      sortMethod: (a, b) => sortDOB(a, b),
```
wmitchell-cambria: nit: We could do `sortMethod: sortDOB,` since the parameters match.

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.test.js
```diff
@@ -43,4 +43,26 @@ describe(' ClientSocialWorkerCardTempate ', () => {
     expect(accessorStatus).toEqual('No prior CANS')
     expect(accessorReminderDate).toEqual('10/26/2018')
   })
+
+  describe('sort method', () => {
+    describe('dob A is more recent than dob B', () => {
```
wmitchell-cambria: We should add a few more test cases here. The examples used would also sort lexically in the correct order, so they don't serve as a very good test of date functionality.

Some examples could be:
* `01/01/2018` and `12/12/2017`
* `03/05/2018` and `04/04/2017`
* Will empty values be sorted to the top of the list or the bottom of the list?

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
```diff
@@ -8,6 +8,12 @@ export function renderDate(datetime) {
   return !datetime || datetime === null ? null : isoToLocalDate(datetime)
 }
 
+export function sortDOB(a, b) {
```
wmitchell-cambria: Let's add the unit tests to `ClientSocialWorkerCardTemplateHelper.test.js`. That way, they test the functionality close to the implementation. If we refactor to use the same sorting method elsewhere, it will be easier to bring the tests along with it.

(A good rule of thumb is:  If you export a new function, it should have associated tests.)

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
KShresthaCwds: Agreed.

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.test.js
KShresthaCwds: Added more test cases.

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
KShresthaCwds: Added test for ClientSocialWorkerCardTemplateHelper.test.js

PR #558, opened 2019-03-04
Refactor: new assessment to be created on the back end

denys-davydov:
## Description
This will integrate FE app with a new rest operation **/clients/:id/assessments/_initialize** added to the BE by https://github.com/ca-cwds/cans-api/pull/231 that will return a new prepared assessment for a client.
This refactoring will help us with implementing new features like reassessment. As yet it already helps us to have a correct assessment creation timestamp, which will help us with statistics. 

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Assessment.service.js
```diff
@@ -7,8 +7,8 @@ export class AssessmentService {
   static update(id, assessment) {
     return apiPut(`/assessments/${id}`, assessment)
   }
-  static fetchNewAssessment() {
-    return apiGet('/instruments/1')
```
wmitchell-cambria: One thing to note here is that we are no longer using the instrument id at all. That is, instrument 1 is baked into the URL. Down the road, if we want to support multiple instruments, we will have to refactor this to support it.

I don't think this is a problem, per se, because we have other functionality that was already moving us in that direction (domain codes in Assessment Summary and Comparisons, reassessment assuming past codes carry over, etc). I just want to draw attention to it so that more developers are aware.

app/javascript/Application/components/Assessment/AssessmentContainer.js
```diff
@@ -135,33 +134,17 @@ export default class AssessmentContainer extends Component {
...
+      const assessment = await AssessmentService.initializeAssessment(this.props.client.identifier)
+      return this.onFetchNewAssessmentSuccess(assessment)
     } catch (e) {
       return this.setState({ assessmentServiceStatus: LoadingState.error })
     }
   }
 
-  onFetchNewAssessmentSuccess(instrument) {
-    const client = this.props.client
-    const assessment = {
```
wmitchell-cambria: Sweet.

app/javascript/Application/components/Assessment/Assessment.service.js
denys-davydov: Thank you for raising it. To me, it was a regular refactoring procedure: the code is not used - the code is gone. As yet FE is not using **/instruments** endpoint at all and probably will not use it again.

denys-davydov:
@wmitchell-cambria I've missed your overall review comment, so I've created a cleanup PR to address it: https://github.com/ca-cwds/cans/pull/564

PR #560, opened 2019-03-05
Patch: Fixing Comparison component render warning.

zliuwd:
## Description

There are couple warning shown when Comparison components were rendered.
screen shot 2019-03-04 at 1 18 12 pm

The reason is used empty `<div/>` as child of LoadingBoundary in ClientAssessmentStatistics, so got some refactoring and resolved it.

## Tests

## Types of changes

## Checklist:


zliuwd:
retest it please

wmitchell-cambria:
I have mixed feelings about both approaches.

The current approach in this PR adds a prop to a couple components that don't feel like they really need it. Why render a Comparison component with no comparison? Why does the AssessmentHistory need knowledge of a comparison at all? (mostly a naming issue)

Denys's proposed approach is cleaner code, but it will cause the loading boundaries to remount, leading to more data fetching. If we want the data to refresh every time you toggle and avoid fetching the comparison when the page loads, this is good...but I think we should be conscious about those tradeoffs rather than having it as a side effect of these warnings.

If I recall, the `<div />` within these loading boundaries was at one point a `Fragment` instead. Did that run into some issue? Maybe the real answer is that the child within the loading boundary should be `isComparisonShown ? <AssessmentComparison /> : null`, and the LoadingBoundary should be set up to handle a `null` child. (Right now, it calls `React.cloneElement` unconditionally, which probably crashes on null.)

denys-davydov:
@wmitchell-cambria Agree about redundant remounting, it is not the best solution. If we change `<div />` with `null` it will fail. But `<Fragment />` instead of `<div />` works great. I think It is the easiest solution.

zliuwd:
### Firstly, thanks for the review from everyone.
#### -For Denys,
The code you provide is very similar as my first version, of course much prettier than mine. That version has an issue: "the page will be reload again" every time when we switch back to history. Then the user experience is not smooth. After talked with Leonid we decide to find a way to make sure the data only load one time for both component. Additionally, I am also not feeling comfortable to create only one boundary for both of them with three fetching at same time. Particularly if we do so, history and comparison's big boundary will lose its reusability. So we wrote the code like that, let them fetch data separately in a same container

zliuwd:
Ok, let we switch back to fragment again.

PR #570, opened 2019-03-11
FIT-404 Apply the new shared library for code quality checks

mariamcws:
## Description
Using updated static analyzer in shared library for code quality checks to check for package.json instead of .eslintrc to address all types of javascript projects

And, changes to apply the new shared library for code quality checks

## Jira Issue link

[Apply the new shared library for code quality checks](https://osi-cwds.atlassian.net/browse/FIT-404)

## Tests

## Types of changes

## Checklist:


Jenkinsfile
```diff
@@ -140,21 +140,20 @@ def incrementTag() {
...
     unitTestStage(container)
   }
 }
 
-def lintStage(container) {
+def lintStage() {
   stage('Lint') {
-    sh "docker exec -t ${container.id} yarn lint"
-    sh "docker exec -t ${container.id} rubocop"
+    lint()
```
edgibbs: Seems fine, but we still might want to allow them to pass in an already built container, to save time in some cases?  It would be a different story though, so this is fine.

Jenkinsfile
```diff
@@ -140,21 +140,20 @@ def incrementTag() {
 
 def buildDockerImageStage() {
   stage('Build Docker Image') {
-    app = docker.build("cwds/cans:${env.BUILD_ID}", "-f docker/web/Dockerfile .")
+    app = buildDockerImageForTest('./docker/web/Dockerfile')
```
denys-davydov: Does it mean that you already have an image name and tag (`"cwds/cans:${env.BUILD_ID}"`) in `buildDockerImageForTest()` function? Will the function work with any other project?

mariamcws: Here are the docs for the function buildDockerImageForTest(path) , it creates a Docker image based on the Dockerfile that is provided by path
https://github.com/ca-cwds/jenkins-pipeline-utils/blob/master/docs/buildDockerImageForTest.md

Jenkinsfile
mariamcws: ok

PR #575, opened 2019-03-13
CANS-53 Comparison graph & CANS-1081 Comparison table upgrade

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-53
https://osi-cwds.atlassian.net/browse/CANS-1081

## Description

- New feature: apply the POC of Comparison graph to actual app
- Upgrade and refactor the Comparison table
## Tests

## Test Coverage
Screen Shot 2019-03-12 at 8 22 13 PM

## Regression Test
After talking with PO, QA and Team, we decided not to include regression test in this PR and story. And we will have a separate story for it.

## Types of changes

## Checklist:


zliuwd:
retest this please

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
```diff
@@ -13,9 +13,10 @@ class AssessmentComparisonLoadingBoundary extends React.Component {
         id: propsId,
         fetch: () =>
           Promise.all([
-            ClientService.getAssessmentComparison(propsId),
+            ClientService.getAssessmentComparison(propsId, true),
+            ClientService.getAssessmentComparison(propsId, false),
             I18nService.fetchByInstrumentId(instrumentId),
-          ]).then(([data, i18n]) => ({ data, i18n })),
+          ]).then(([underSix, aboveSix, i18n]) => ({ data: { underSix, aboveSix }, i18n })),
```
zliuwd: ### Sharing some offline talking with Leonid, and this will let the reviewers know why we want to have 2 times getAssessmentComparison fetching:
- #### The problem we want to resolve: 
   How to decide that should render the "age range switch buttons" or not?  Because the AC required that if a client only have data in one age range then we should hide the "age range switch buttons".

- #### How the new API works:
  - First: When pass isUnderSix params to API: it will returns the data for the specific age range.
  - Second: When does not pass isUnderSix params to API: it will returns the data which belong to the age range from recently completed assessment. This functionality keep our old comparison working during the upgrade in these days.
- #### Could we use the default API return to identify "should render age switch or not" ?
No, we still don't know if the client has data or not in another age range, even though we got his recently complete assessments. By the default API we still need some logic to identify what age range this default return is, then we need SECOND fetching OUT OF THE BOUNDARY to make sure if the client has data in another age range. 
- #### Could Client History help?
No, Client history API return does NOT have age range flags. It means we will not know if the client has data in another age range or not by both of History API returns and Comparison API default returns.

- ### The choice: 
"All in all we have to fetch getAssessmentComparison twice" but we have choice:
- First: Fetching both sides in beginning 
  - Benefits:
    1.keep all fetching in Boundary
    2.It will be very smooth when we switch between age range, no any extra fetching after that
  - Drawback:
    One more fetching in the boundary

- Second: Do first fetch in boundary then do second fetch in some where
  - Benefits: NO obvious benefits
  - Drawback: 
    1. reloading every time when switch between the age
    2. spread the fetching to some where out the boundary

wmitchell-cambria:
I will try to take a look at this today. Sounds like there was some interesting discussion happening. Thanks for posting a summary.

zliuwd:
Thanks Will!

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
```diff
@@ -1,40 +1,78 @@
...
+import ComparisonAgeSwitchButtonGroup from './ComparisonAgeSwitchButtonGroup'
 
-const AssessmentComparison = props => {
-  const { comparisonRecords } = props
-  if (!comparisonRecords || !comparisonRecords.data || !comparisonRecords.i18n) {
-    return null
+class AssessmentComparison extends React.Component {
+  state = {}
+
+  static getDerivedStateFromProps({ comparisonRecords }) {
```
wmitchell-cambria: It looks like this component can be written without using `getDerivedStateFromProps`. As is, it is [unconditionally copying information from props to state](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state), which is an anti-pattern. Most of the state information can be calculated from props directly during render. You may need a state for `currentDataKey` still, which would be set in the constructor and updated by `handleAgeSwitch`.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
wmitchell-cambria: I think that explanation sounds reasonable (with the caveat that I am not super familiar with the data involved, so I trust y'all). I do want to point out that most browsers limit the page to 2 in-flight ajax calls simultaneously, so adding a third call here may result in a noticeable lag in loading of all three requests. It may still be the best option, but I wanted to make sure it was something you were considering.

app/javascript/Application/components/Client/AssessmentComparison/comparisonCommonHelper.js
```diff
@@ -0,0 +1,22 @@
+export const dataAnalyzer = data => {
```
wmitchell-cambria: If this is only used in the single component, can it be included there directly instead of creating a helper file? That has a few benefits:
* It would be included in the same 'unit', and tested directly through the component props, resulting in better test coverage.
* All relevant information for the component will be in the same file, making it more encapsulated and easier to read later.
* All of the logic will be in the same file, which means CodeClimate can keep us honest about the file size. (Splitting it up early gives us twice the line limit.)

app/javascript/Application/components/Client/AssessmentComparison/comparisonCommonHelper.js
```diff
@@ -0,0 +1,22 @@
...
+  if (!data) {
+    return { currentDataKey: '', showSwitch: false }
+  }
+  const result = { currentDataKey: '', showSwitch: true }
+  if (data.underSix.event_dates.length < 1) {
```
wmitchell-cambria: The PropTypes for `data.underSix` in the comparison component declared it as an optional key, and as a generic object. That means that `data.underSix` could be undefined, `data.underSix.event_dates` could be undefined, and `data.underSix.event_dates.length` could be undefined. All of these would give unexpected results (crash, crash, false).

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/ComparisonGraph.js
```diff
@@ -0,0 +1,43 @@
...
+import ComparisonGraphInner from './ComparisonGraphInner'
+import { graphColors, datesGenerator, domainBarsDataGenerator } from './comparisonGraphHelper'
+
+const colors = graphColors
+
+class ComparisonGraph extends React.PureComponent {
+  nextColorIndex = 0
+
+  nextColor = () => {
+    const number = this.nextColorIndex++ % colors.length
```
wmitchell-cambria: Please find a way to write this without using stateful, non-React properties on the class. For example, we could either pass the list of colors to the `ComparisonGraphInner` or have `ComparisonGraphInner` import the list of colors directly and use it when mapping over the domains.

Properties like this can lead to inconsistent behavior, as they operate outside of the React lifecycle. Additional renders could change the color of the graph (and React does not _guarantee_ that PureComponent will not perform unnecessary renders), or using a separate instance of the ComparisonGraph that renders the same React elements will behave differently than using the same instance twice.

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/comparisonGraphHelper.js
```diff
@@ -0,0 +1,60 @@
...
+import { getI18nByCode } from '../../../common/I18nHelper'
+
+export const graphColors = ['#7bb5ea', '#414146', '#93ed86', '#f7a564', '#808080']
+
+const dateInfoGenerator = originalDates => {
+  return originalDates.reduce((result, date) => {
+    const id = date.assessment_id
+    const dateKey = isoToLocalDate(date.event_date)
+    const keys = [dateKey, dateKey]
+    // keys[0] is startKey, keys[1] is fixedKey, write like this for passing code climate checking
```
wmitchell-cambria: What was the issue with CodeClimate?

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/comparisonGraphHelper.test.js
```diff
@@ -0,0 +1,159 @@
...
+        type: 'REGULAR',
+        assessment_id: 65006,
+      },
+      {
+        assessment_id: 65007,
+      },
+    ],
+  },
+]
+describe('datesGenerator', () => {
```
wmitchell-cambria: Similar to my comment above. If these helper methods are only used in one file, and that file is well within our CodeClimate file size limit, I think we should keep these together with the component. Helper files tend to be "bag of function" files without a focused purpose. Keeping the functions as close to their component as possible helps with this.

zliuwd:
After talking with PO, QA and Team, we decided not to include regression test in this PR and story. And we will have a separate story for it. Thanks!

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparisonLoadingBoundary.js
zliuwd: Yes, it just what I concerned before -- this triple fetching stuff. But as you said, it probably the best solution I could found with current API.

zliuwd: But it woks fine because the return data size is not so large.

app/javascript/Application/components/Client/AssessmentComparison/AssessmentComparison.js
zliuwd: Thanks for mention that, learned from this case

app/javascript/Application/components/Client/AssessmentComparison/comparisonCommonHelper.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/comparisonGraphHelper.js
zliuwd: The issue is: parameters amount of one function can not greater than 4. 
The repeatChecker function down below has 5 parameters before, so here I have to combine the keys to satisfy this requirement

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/ComparisonGraph.js
zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonCommonHelper.js
zliuwd: You right, actually data.underSix or data.aboveSix will always defined because API will return {} for them if there is no data, so they should be required. Done.

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/comparisonGraphHelper.test.js
zliuwd: I partly agree with this point, but I do not agree that "If these helper methods are only used in one file, and that file is well within our CodeClimate file size limit" should be the only condition or strict condition for keeping the function in component.
**So I changed your last request and I would like to keep this one as Is**, the reasons down below:

For some complex function and  which only use for transform data structure, I prefer to keep them separately in helpers, this will have at least two benefits:
1. could make sure this function be fully and much easily tested
2. could keep the component with more clear structure.
**But attention that what I said is only for the pure data transform functions**, for example:
this 'datesGenerator', actually it took 20 more lines to refine and transform dates from the raw data 'graphDates', but  'ComparisonGraph' only need to know it pass a dates to 'ComparisonGraphInner'. In this case put that transform function in helper will making the test more efficient and effective. At same time there will be no confuse for the devs who will work on this component later, because it just is 'date to date' without any hide trick.

wmitchell-cambria: Sounds good to me. In that case, can we name the file something like `ComparisonDataTransformations.js`? My biggest issue with "Helper" is that it doesn't describe what the file does, so it attracts a random mismash of unrelated functions.

zliuwd: Sure I have the same feeling with that name issue, thanks for mentioned.

zliuwd: done

app/javascript/Application/components/Client/AssessmentComparison/comparisonGraph/ComparisonGraphInner.js
```diff
@@ -3,7 +3,20 @@ import PropTypes from 'prop-types'
 import { BarChart, CartesianGrid, XAxis, YAxis, Tooltip, Legend, Bar, Cell, ResponsiveContainer } from 'recharts'
 import { graphColors } from './comparisonGraphHelper'
 
-const ComparisonGraphInner = ({ domainBarsData, dates, nextColor }) => {
+export const backupColor = () => {
+  // this is a backup for the accident case that graphColors length shorter than amount of assessments
```
zliuwd: I also have test to making sure the assessments amount should less than graphColors array length. This backup just want to make sure it much more safe.

zliuwd: It means that even we change the Max required assessments amount , or delete some color by mistake, the app will not crash, then devs could tell this by test fail

legal/licenses.csv
```diff
@@ -255,7 +255,7 @@ buffer-xor,1.0.3,MIT,https://github.com/crypto-browserify/buffer-xor
 builder,3.2.3,MIT,http://onestepback.org
 builtin-modules,1.1.1,MIT,sindresorhus.com
 builtin-status-codes,3.0.0,MIT,bendrucker.me
-bundler,1.17.1,MIT,http://bundler.io
+bundler,1.16.4,MIT,http://bundler.io
```
LeonidMarushevskyi: Why bundler version went back?

PR #577, opened 2019-03-14
CANS-1078. Reassessment data is carried over

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1078

## Description
The data from the preceding assessment is carried over to the new reassessment form. Some of the data is cleaned.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentHelper.js
```diff
@@ -243,3 +243,18 @@ export const handlePrintButtonEnabled = state => {
 export function isSubsequentType(type) {
   return AssessmentType.subsequent === type
 }
+
+export function preparePrecedingAssessment(precedingAssessment, eventDate) {
+  precedingAssessment.id = null
```
wmitchell-cambria: I think this is the function that was standing out to me in the previous review. It mutates the `precedingAssessment` object, and we should be very careful with impure functions like this. Can we please refactor it? We can either:
1. return a new object (leaving the parameter object untouched),
2. or have it return undefined (so that the caller is forced to reuse the same object afterwards, drawing attention to the mutation).

denys-davydov: Good hint, here is a PR to address it with the second option: https://github.com/ca-cwds/cans/pull/584

PR #579, opened 2019-03-15
CANS-838-AddCANSButtonTextToggle

AK56:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-838
## Description

The 'ADD CANS' link is switched with a button desired in the mockup with the text that either says 'Add CANS' if the client never had a CANS completed before or the button text 'Add reassessment' if the client previously had a CANS completed before with a valid preceding_assessment_id. I am adding a preceding_assessment_id check to ensure the best User experience i.e. the user expects the reassessment to be pre filled with item ratings from that of preceding_assessment if user choose to 'Use previous rating' from the modal after clicking the button 'Add reassessment'
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentHelper.js
```diff
@@ -243,3 +243,10 @@ export const handlePrintButtonEnabled = state => {
 export function isSubsequentType(type) {
   return AssessmentType.subsequent === type
 }
+
+export const hasOneCompletedForReassessment = assessments => {
+  if (!assessments || assessments.length === 0) return false
+  return (
+    assessments.filter(assessment => assessment.status === 'COMPLETED' && assessment.preceding_assessment_id).length > 0
```
denys-davydov: In requirements it is said that previous completed assessment should be for the same Case/Referral number. I think we should add this validation here instead of `assessment.preceding_assessment_id ` which shouldn't be used for this logic.

AK56: Great call and catch! Had some additional offline chat with @denys-davydov as well, so just pushed the desired changes

PR #581, opened 2019-03-18
Cans 844 Review Reassessment

abelayt:
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -195,6 +196,7 @@ Assessment.propTypes = {
   handleWarningShow: PropTypes.func,
   i18n: PropTypes.object.isRequired,
   isDefaultExpanded: PropTypes.bool,
+  isReviewed: PropTypes.bool,
```
AK56: This seems to be a prop not being set by the parent component on this component. But since the domain's `isReviewed` is driven by domain's 'is_reviewed' flag, perhaps we can use that here to declare `isReviewed` and pass it along as the prop to `Domain` component

Also no test coverage for the `isReviewed` here

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -40,6 +42,13 @@ class Domain extends Component {
...
+  handleOpenToReview = event => {
+    this.props.domain.is_reviewed = true
+    this.setState({
+      isReviewed: true,
+    })
+  }
+
   handleExpandedChange = event => {
     this.setState({
       expanded: !this.state.expanded,
```
AK56: We can make a call to `handleOpenToReview` here to fix the bug

app/javascript/Application/components/Assessment/Assessment.js
abelayt: Thanks! I was not sure whether we need to keep Expand/Collapse All and was planning to discuss this during our standup today. I remember discussing not to have it until all domains are opened to review but I do not see it on A/C. In any case I've removed the Open to review button when Expand All is clicked.

Also removed isReviewed from Assessment for now. I do not see its use so far in this story, but if needed let me know. Thanks.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -40,9 +42,18 @@ class Domain extends Component {
     return null
   }
 
+  handleOpenToReview = event => {
+    this.props.domain.is_reviewed = true
```
wmitchell-cambria: This mutation should be avoided. Modifications to the `domain` prop should be handled via a callback function. This is for several reasons.
1. There are situations where React will only perform a shallow comparison of props or state values. In this case, the mutated `domain` is the same reference as the original `domain`. This can cause issues when using PureComponent (even elsewhere in the React tree), among other things.
2. Using a callback to modify the domain in the parent allows there to be a single source of truth for `is_reviewed`. That source of truth is the domain itself. By using a callback that modifies the domain in the ancestry, an update cycle will be triggered that passes the new value down as a prop. Tracking the `isReviewed` in state here becomes redundant, allowing us to simplify the component.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -188,6 +216,7 @@ Domain.propTypes = {
   index: PropTypes.number.isRequired,
   isAssessmentUnderSix: PropTypes.bool,
   isDefaultExpanded: PropTypes.bool,
+  isReviewed: PropTypes.bool,
```
wmitchell-cambria: Is this prop used?

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -121,13 +121,32 @@ $item-confidential-label-height: 6.05rem
 //********************
 //Domain
 //********************
+.review-regular-button, .review-regular-button:hover:not(:disabled)
+  border-color: $dark-cyan
```
wmitchell-cambria: If this button style does not exist in the Component Library, we should use a button style that does. We should not override 'primary' to look custom.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -111,15 +122,32 @@ class Domain extends Component {
...
     const totalScore = totalScoreCalculation(validItems)
     const warningText = <span className={'caregiver-warning-text'}> Caregiver Name is required</span>
+
     return shouldDomainBeRendered(isAssessmentUnderSix, domain) ? (
       <Fragment>
         <ExpansionPanel expanded={expanded} onChange={this.handleExpandedChange} elevation={0}>
           <ExpansionPanelSummary
             expandIcon={
-              <ExpandMoreIcon id={`domain${index}-expand`} style={{ height: '28px', color: '#000000', padding: '0' }} />
+              isReviewed || expanded ? (
```
wmitchell-cambria: Rather than inlining a ternary statement here, consider creating a new component that encapsulates this behavior. That will help keep `Domain.js` under the file size limit and make it easier to test these different conditional scenarios in its unit tests without having to provide lots of props.

Also, I suspect that one reason the regression tests are breaking is that the expand icon is used in the tests to expand the domains. If the icon is not there, it cannot expand them. A good workaround may be to always render the icon alongside the button (and wait for a separate story to move the chevron on the left side of the domain header).

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -138,6 +138,16 @@ class Assessment extends Component {
     this.props.onAssessmentUpdate(assessment)
   }
 
+  updateDomainIsReviewed = (code, caregiverIndex) => {
+    const assessment = clone(this.props.assessment)
+    assessment.state.domains.map(domain => {
```
wmitchell-cambria: nit: `forEach` here instead of `map`. We don't care about the return value.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -121,13 +121,38 @@ $item-confidential-label-height: 6.05rem
...
   flexWrap: nowrap
   flexDirection: row
   justifyContent: center
   padding: 0 1.25rem 0 0
 
+.domain-metric-with-review
+  display: flex
+  flexWrap: nowrap
+  flexDirection: row
+  justifyContent: center
```
wmitchell-cambria: These properties should be hyphenated, not camel case.

wmitchell-cambria:
I chatted with @abelayt offline. Re: the Domain expandIcon refactoring, he pointed out that the component is using a lot of Material UI components and will soon be changing to use new designs anyways. He suggested that refactoring it now doesn't make sense given that upcoming work, and I agree.

For the outline button, the big concern is that we will have to carry around these custom styles indefinitely. To address this, I created a story to switch over to the CL outline styles once they are implemented: https://osi-cwds.atlassian.net/browse/CANS-1120 In the future, we should try to limit ourselves to the buttons in the Component Library. If those buttons are not ready, we should choose suitable alternatives that are ready, and then update them later. (In other words, avoid custom styles!)

PR #596, opened 2019-03-25
CANS-1000: Complete modal window reworked

LeonidMarushevskyi:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1000

## Description

Complete warning modal replaced with cwds/components Modal component. Modal text and buttons behaviour changed. Using styles from Component Library.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/CompleteModal.js
```diff
@@ -0,0 +1,51 @@
...
+import { Modal, ModalBody, ModalHeader, CardTitle, Button } from '@cwds/components'
+
+const CompleteModal = ({
+  isCompleteModalShown,
+  handleCompleteWarning,
+  handleCompleteAssessment,
+  handleSaveAssessment,
+}) => (
+  <div>
+    {isCompleteModalShown ? (
```
denys-davydov: We probably do not need the `div` wrapper and `isCompleteModalShown` ternary here.  I believe the `Modal`'s `isOpen` prop will handle it for us. Please try to remove them and use this line for the modal:
`<Modal isOpen={isCompleteModalShown} size="lg">`

LeonidMarushevskyi: Good catch. Done

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -497,16 +497,18 @@ def click_save_and_summary_card_is_shown
   def warning_and_summary_card_shown_after_complete_button_clicked(is_reassessment)
     expect(@form.header.authorization_radio_no.checked?).to be(true)
     @form.footer.complete_button.click
-    expect(@form.app_globals.complete_warning_modal['style']).to eq('display: block;')
-    @form.app_globals.cancel_button_of_warning.click
+    expect(@form.app_globals).to have_complete_warning_modal
+    sleep(2)
```
nasar-cambria: @zliuwd created a method for sleep and focus. Was that deleted some how I see you are using sleep here?

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -497,16 +497,18 @@ def click_save_and_summary_card_is_shown
...
+    @form.app_globals.complete_warning_save_return_button.click
     if is_reassessment
       expect(@form.global).to have_reassessment_page_header
     else
       expect(@form.global).to have_assessment_page_header
     end
     expect(@form.footer).to have_complete_button
     @form.footer.complete_button.click
-    @form.app_globals.agree_button_of_warning.click
+    sleep(2)
```
nasar-cambria: Same comment as above

spec/regression/case_worker_scenarios_spec.rb
LeonidMarushevskyi: The problem here is not related to scrolling. For some reason, Selenium has an issue with clicking buttons in modal windows from the component library. The temporary solution is explicitly to wait until the modal window will be completely rendered. Will continue looking for a better solution.

PR #597, opened 2019-03-25
CANS-1090 Tooltips don't Disappear

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-1090

## Description
During the scroll function, the tooltips don't disappear after you're no longer hovering your mouse cursor over them. Updating tooltips to be used from Component library which fixes the issue.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -152,10 +151,11 @@ class Domain extends Component {
               <Grid item xs={8}>
                 <Typography variant="title" style={{ color: '#0e6f89', fontSize: 16 }}>
                   {title}
+                  <Icon className="domain-help-icon" icon="info-circle" id={`domain-${index}`} />
                   {description ? (
-                    <Tooltip title={description} placement="top" classes={{ tooltip: 'tooltip_' }}>
-                      <Icon className="domain-help-icon" icon="info-circle" />
-                    </Tooltip>
+                    <UncontrolledTooltip style={{ minWidth: '20rem' }} target={`domain-${index}`} placement="top">
```
LeonidMarushevskyi: Can we move style to sass file?

nasar-cambria: It doesn't take className props. So, I have to use inline styles.

denys-davydov: @nasar-cambria Try to wrap `{description}` with a `<div className="class-name-with-min-width">` to avoid inline usage of styles.

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -152,10 +151,11 @@ class Domain extends Component {
               <Grid item xs={8}>
                 <Typography variant="title" style={{ color: '#0e6f89', fontSize: 16 }}>
                   {title}
+                  <Icon className="domain-help-icon" icon="info-circle" id={`domain-${index}`} />
```
denys-davydov: Not sure if the `<Icon>` should be rendered when there is no `description`. It should probably have the same rendering rules as the following `<UncontrolledTooltip>` element

app/javascript/Application/components/Assessment/Domain.js
nasar-cambria: @denys-davydov good thought but we need to increase the width of the tooltip and by increasing the width of description we may end up cropping some part of it. EX: 
Screen Shot 2019-03-25 at 11 02 54 AM

LeonidMarushevskyi:
Bug created to team CL https://osi-cwds.atlassian.net/browse/CL-147

app/javascript/Application/components/Assessment/Domain.js
wmitchell-cambria: Shouldn't the width of the tooltip automatically adjust based on the content? To me, that sounds like the underlying problem here. We should only have to apply a className to the tooltip in very rare situations, now that most of the styling and behavior is owned by the Component Library.

PR #616, opened 2019-04-01
Regression test suite fix

denys-davydov:
## Description
Regression small fix

## Tests

## Types of changes

## Checklist:


spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -245,7 +245,6 @@ def scroll_to_top
 
   def validate_new_assessment(current_date, client_name)
     @client_profile.go_to_recently_updated_assessment(current_date)
-    expect(@form.global).to have_assessment_page_header
```
denys-davydov: This is to fix regression suite in integration

wmitchell-cambria: Why does Integration not have an assessment page header?

denys-davydov: The removed line was previously needed to validate that we are on the Assessment form page after we went to the page. Now the line is not needed because we are also validating that we are on the Assessment form page with the next line:
```expect(@form.header.child_name).to have_content(client_name)```
The issue with the removed line is that it doesn't take into account the fact that the page could be a Reassessment form as well as an Assessment form. If the validation made by this line is still made by the next line (which also validates the child name), then it can be removed.

wmitchell-cambria: Given we are using curated test data, I would expect the header to be the same in any environment. That is, if it is a Reassessment in Integration, why is not a Reassessment in Preint or local? That sounds like bad data to me, which we should fix.

denys-davydov: Yes, forgot to mention, that the reason for the fix is that the test is failing in Integration while it is not failing elsewhere (yet).
The reason for the failure is (1) the data in Integration and (2) the manner the test is written with. The data (1) could be entered manually and finding a root cause for its existence might be time consumptive. The manner the test is written with (2) could be changed easier because for this particular test case it doesn't actually matter if it is an Assessment form or a Reassessment form. So removing a redundant line seems to be a better solution because: (a) the test becomes less dependent on things that are not under test here; (b) the suite becomes more stable and independent from data entered manually; (c) we are removing redundancy; (d) it is cheaper.

denys-davydov: However, we can still investigate the data issue

wmitchell-cambria: We should definitely investigate. Making the test more stable in the face of changing data goes hand in hand with widening the set of allowed behaviors that get through the test. Because we rely on these tests to help us know when something is going wrong, they should stay as tightly focused in what they test as possible. If we are getting bad data manually entered which affects our tests, that is a larger problem that we should also address.

An alternative way to look at the investigation is not as a more expensive approach, but instead as the test indicating a valid issue. When tests fail, changing the expectations of the test should be a last resort.

denys-davydov: With removing the line test expectation is not changed. It is described here https://github.com/ca-cwds/cans/pull/616#discussion_r271446194

denys-davydov: And here https://github.com/ca-cwds/cans/pull/616#discussion_r271455683
>  for this particular test case it doesn't actually matter if it is an Assessment form or a Reassessment form

wmitchell-cambria: It _does_ change the expectation. The line removed is literally an expectation. Removing it allows both Assessments and Reassessments pass the test, whereas before only Assessments could. We can debate whether this is an appropriate change to the test expectations, but it is a _fact_ that it is a changed expectation.

denys-davydov: It _does not_ change the test expectation, here is why.
The purpose of the line was to ensure that the user was routed to the assessment form page and the page is loaded, and now it is ensured by the next line, which by time was written 2 weeks after it to ensure that the *client data* is loaded as well (not only the page title, because the test was sometimes failing with long data loading), so the previous line became redundant (and there were no Reassessment functionality at that time), but it was not removed at that time by mistake. The test or the line was never created to test 'CANS Assessment/Reassessment Form' label or to test the difference between Assessment/Reassessments behavior, which was not and still is not important for this test scenario. The scenario is called 'Case worker creates and deletes assessment' and it tests the *deletion* because the creation is tested in many other scenarios, and for the deletion it doesn't matter if it is an assessment or reassessment. Additional indirect sign that it doesn't matter is that there is no 'Case worker creates and deletes *re*assessment' scenario. 

So, the expectation of the test scenario is that an assessment form can be deleted, and the expectation was not changed. The removed line was there to ensure that a page is loaded, it is still ensured by the next line that is still there.

denys-davydov:
Reverted the refactoring for reassessment form preparation to separate the concerns. This code will is moved to a separate PR https://github.com/ca-cwds/cans/pull/618

spec/regression/case_worker_scenarios_spec.rb
abelayt: The scenario the test on Int was failing was on checking the title where it is not the same now for Reassessment. This can be fixed either by fixing the test in Int env which still tries to validate the title which has been changed or the approach this PR is taken to fix. I still approve this PR because this approach still makes sense to me.

denys-davydov: Need to mention that we are not talking about **test data**, we are talking about currently existent data on the environment that was created in any possible way, by our tests or by users or by direct db manipulation. The data is not corrupted or "broken" from any perspective, it is just "like real" data. With removing the line, this test scenario becomes independent from the currently existent data in the environment and can be successfully run against any preexistent data on the environment. If we do not remove the line, we will have to "fix" or "prepare" already existent environment data manually from time to time, which regular action I believe we want to avoid. Removing the line will allow both Assessments and Reassessments to pass the test scenario because the type of the assessment is not a concern of this test scenario.

wmitchell-cambria: OK, let me recap, because I have already addressed these points:

1. We are relying on curated data in each environment. We expect the same results in each environment. If we are getting different results, that is a failure of our regression tests that should not be fixed by changing the regression tests. If our data is no longer properly curated, we should fix it and make an effort to prevent it from changing again in the future.
2. Prior to this change, we expected the created assessment in Integration to be an initial assessment, not a reassessment.
3. The test was written to only allow assessments to pass it. The method name references new assessments. It verifies that the header describes a new assessment. The corresponding cell in the spreadsheet references the "Add CANS" button.
4. Whether or not this expectation was originally intended, modifying the test by removing the line changes this expectation.

I am not going to remove my block on this one based on the discussion we have had so far. Regression tests have no value if we change them to match unexpected differences between environments.

denys-davydov:
I could recap my point of view in one simple argument: **assessment/reassessment distinguishing issue is not a concern of this particular test scenario**. That is enough for me to drop that line of code with a clear conscience. 
If it is not enough for the team, I am fine with closing this PR. The data in Integration was fixed manually yesterday, so the test scenario will not fail there for some period of time. Feel free to create a Tech Debt story that will cover the issue if you think it is needed or if the issue arises again.

PR #620, opened 2019-04-03
CANS-1048 - Assessment option radio groups

wmitchell-cambria:
## JIRA

[CANS-1048 - Redesign Top Card on Assessment - Caregiver and Auth for Release](https://osi-cwds.atlassian.net/browse/CANS-1048)

## Description

Behaviorally, this is a style change of the radio option groups in the Assessment Form Header. All user functionality should be the same. The radio groups were placed into boxes with helpful messages inside them. To support this, some refactoring was done to share code between the two very similar sets of options.

Note that this is not yet using Component Library radio buttons, as those will only be released next sprint at the earliest. In the meantime, we continue using Material UI.

Also note that I was out of office for a significant time in the middle of this work, so please review carefully in case I missed some cleanup. :-/

## Tests

I updated the selectors for the new radio button styles. I ran some regression tests locally, but I expect to use Jenkins to help with the full suite run again. I manually tested across browsers. A pretty large screen size is required for sane line breaks (~1200px).

## Types of changes

## Checklist:

Screen Shot 2019-04-03 at 2 49 40 PM


wmitchell-cambria:
Retest this please

app/javascript/Application/components/Assessment/AssessmentFormHeader/RadioGroupMessageBox.test.js
```diff
@@ -0,0 +1,77 @@
...
+    const col = render().find('Col.radio-group-box')
+    expect(col.exists()).toBe(true)
+  })
+
+  it('renders a Radio Group', () => {
+    const wrapper = render({ legend: 'My Group' })
+    expect(findRadioGroup(wrapper).props().legend).toBe('My Group')
+  })
+
+  it('passes props down to RadioGroup', () => {
```
zliuwd: Really a good idea to check the props.

app/javascript/Application/components/Assessment/AssessmentFormHeader/AssessmentOptions.js
```diff
@@ -0,0 +1,61 @@
...
+  onHasCaregiverChange: PropTypes.func.isRequired,
+  onHasCaregiverNoClicked: PropTypes.func.isRequired,
+  substanceUseItemsIds: PropTypes.shape({
+    underSix: PropTypes.array.isRequired,
+    aboveSix: PropTypes.array.isRequired,
+  }).isRequired,
+}
+
+AssessmentOptions.defaultProps = {
+  isDisabled: undefined,
```
zliuwd: isDisable is from AssessmentContainer.js state "isEditable", and AssessmentContainer.js has default state value be set as "false", is there any chance it could be undefined? (of course, it was set to undefined in AssessmentContainerInner.js default props also...)

wmitchell-cambria: I'm not totally sure. I think there are a lot of cases here where we account for more states than we need to. If we get stricter about what props can be undefined and which are required, it could simplify a lot of checks throughout our tree.

wmitchell-cambria:
@denys-davydov - I agree. I just pushed a commit that fixed up the CanReleaseInfoQuestion to take a boolean, similar to the HasCaregiverQuestion. It looks like the assessment should be required at that point in the tree, too. I'll mark it as required. Good catch.

app/javascript/Application/components/Assessment/AssessmentFormHeader/AssessmentOptions.js
wmitchell-cambria: Looking at this a little more, I think it is unlikely that it will ever be undefined in our current tree. However, it does seem like an optional prop at a fundamental level... if nobody says that it is disabled, it will be enabled. If I marked it as required, anybody using the component would explicitly have to say `isDisabled={true}`.

zliuwd: Very nice to know it, thanks for checking 👍

wmitchell-cambria:
Heads up that Amera has requested a little more time to review this one, so we're not in a rush to merge this if it goes green. (On a similar note, I am hoping that this build can finish running acceptance tests before updating to latest master, so I can see the results before the build is aborted.)

wmitchell-cambria:
Retest this please

PR #624, opened 2019-04-05
CANS-831 Add component for Reminder Date Pill

TylerOSI:
## Description
Adding a Reminder Pill for Reassessment date

## Outstanding issues
- minor styling and spacing (&nbsp; used to avoid changing styling in component library)

## Tests

## Types of changes

## Checklist:


TylerOSI:
Jenkins test this please

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
```diff
@@ -20,6 +20,12 @@ export function renderReminderDate(date) {
   return moment().isAfter(dateToStartShowingReminderFrom) ? isoToLocalDate(date) : null
 }
 
+export function withinOneMonth(date) {
```
TylerOSI: this should be within 30 days instead of within one month

app/javascript/Application/components/Client/ReminderDatePill.js
```diff
@@ -0,0 +1,54 @@
...
+const badgeColor = days => {
+  if (days < 0) return 'danger'
+  return 'warning'
+}
+
+export const ReminderDatePill = ({ original: { reminder_date: reminderDate } }) => {
+  if (!reminderDate) return null
+  const date = isoToLocalDate(reminderDate)
+  const now = moment({ hours: 0 })
+  const days = moment(reminderDate, ISO_DATE_FORMAT).diff(now, 'days')
```
TylerOSI: some of these variables should be extracted into helpers so they can be reused in the application.

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
```diff
@@ -20,6 +20,12 @@ export function renderReminderDate(date) {
...
+  if (!date) return false
+  const dateToStartShowingReminderFrom = moment(date, ISO_DATE_FORMAT).subtract(1, 'months')
+  return moment().isAfter(dateToStartShowingReminderFrom) && moment().isBefore(moment(date))
```
wmitchell-cambria: This method is using `isAfter` and 1 month, whereas @denys-davydov has a method using `isSameOrAfter` and 30 days. Please coordinate with him to use a single approach. Multiple implementations lead to edge case bugs.

app/javascript/Application/components/Client/ReminderDatePill.js
```diff
@@ -0,0 +1,54 @@
...
+
+const badgeColor = days => {
+  if (days < 0) return 'danger'
+  return 'warning'
+}
+
+export const ReminderDatePill = ({ original: { reminder_date: reminderDate } }) => {
+  if (!reminderDate) return null
+  const date = isoToLocalDate(reminderDate)
+  const now = moment({ hours: 0 })
```
wmitchell-cambria: Please be sure to double check UTC time on this one. Does it return the expected day if it is after 5pm Pacific?

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
TylerOSI: @denys-davydov implementation is correct. Using one month instead of 30 days here is an oversight on my part.

app/javascript/Application/components/Client/ReminderDatePill.js
TylerOSI: Moment uses the browsers date and time, which essentially uses the operating systems date and time. The main reason for being concerned with timezone is if users of the application are in different time zones or they are traveling. Neither of these situations really apply for cans (unless our PO disagrees). 

The application currently does not concern itself with timezone. If we need to be concerned about timezone, I recommend using [moment-timezone](https://momentjs.com/timezone/) when saving to force the timezone to PST/PDT and when doing other operations after retrieving.

To check this code against UTC time, I would need to change my system time to be UTC -- and that seems like an unlikely real world usage given that the application will always be used in PST/PDT.

app/javascript/Application/components/Client/ReminderDateCell.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Badge } from '@cwds/components'
+import { isoToLocalDate, daysFromNow } from '../../util/dateHelper'
+
+const THIRTY_DAYS = 30
```
denys-davydov: This const is to be reused

app/javascript/Application/util/dateHelper.js
```diff
@@ -129,3 +129,7 @@ export function calculateDateDifferenceInYears(startDate, endDate) {
   const duration = moment.duration(moment(endDate).diff(moment(startDate)))
   return Math.floor(duration.asYears())
 }
+
+export const now = () => moment({ hours: 0 })
```
denys-davydov: This is more like `today` than `now` because it has no time. To be reused.

app/javascript/Application/util/dateHelper.js
```diff
@@ -129,3 +129,7 @@ export function calculateDateDifferenceInYears(startDate, endDate) {
...
+
+export const daysFromNow = date => moment(date, ISO_DATE_FORMAT).diff(now(), 'days')
```
denys-davydov: To be reused

PR #627, opened 2019-04-05
Update bootstrap and rails for security patches

wmitchell-cambria:
## Description

Small updates to Bootstrap and Rails dependencies to address flagged security alerts in GitHub.

## Tests

I didn't really run any tests other than loading up the app. These updates should have no visible effect.

## Types of changes

## Checklist:


wmitchell-cambria:
Looks like the Bootstrap upgrade went bad somewhere. I'll have to investigate further.

wmitchell-cambria:
Here is a quick explanation of what happened with the original update... Bootstrap is required as a direct dependency and also as a dependency of cwds/components. My first attempt was to just update both directly since we were only behind a couple minor versions on the component library. However, a recent version of the component library deprecated a few fields, specifically around buttons, and that caused lots of warning messages to be logged. Warning messages are currently treated as failures in our unit tests (which was our choice, not CL's fault).

The solution for now was to avoid upgrading the component library itself, and only update the underlying Bootstrap dependency for it. The Bootstrap version is compatible with the package.json for the component library, so that worked well. Now we are locked in to the patched Bootstrap version without having to update all of the extra libs.

PR #630, opened 2019-04-08
CANS-1137 Do not carry over N/A checked boxes in Reassessment

KShresthaCwds:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1137
## Description
Visual Indicator for N/A option and default unchecked N/A option for Reassessment
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Item/ItemNaCheckbox.js
```diff
@@ -17,7 +17,9 @@ const PrimCheckbox = withStyles({
 })(Checkbox)
 
 const ItemNaCheckbox = props => {
+  const EIGHT = 8
```
wmitchell-cambria: We have a helper method in ItemHelper.js `isNaRating`, so that we don't have to reference `8` in multiple places. We should use that here. It's already imported.

app/javascript/Application/components/Assessment/ReassessmentHelper.js
```diff
@@ -38,12 +38,22 @@ export const prepareCaregiverDomains = (reassessment, precedingAssessment) => {
 }
 
 export const prepareItemForReassessment = (item, previousRatingAndConfidential) => {
+  const EIGHT = 8
```
wmitchell-cambria: Same here. We should not reference `8` directly outside of the helper.

Also, if we _were_ going to define this as a constant here, we should pick a better name. The reason the magic-number lint rule exists is to encourage developers to choose good names for large constants. `EIGHT` does not tell you anything about _why_ the constant exists.

app/javascript/Application/components/Assessment/ReassessmentHelper.js
```diff
@@ -38,12 +38,22 @@ export const prepareCaregiverDomains = (reassessment, precedingAssessment) => {
...
+    : previousRatingAndConfidential.confidential
   if (previousRatingAndConfidential) {
-    item.rating = previousRatingAndConfidential.rating
-    item.confidential = item.confidential_by_default
-      ? item.confidential_by_default
-      : previousRatingAndConfidential.confidential
+    item.rating = itemRating
+    item.confidential = itemConfidential
   }
+  return null
```
wmitchell-cambria: If the return value is not used, we can return nothing. Just use `return` alone.

app/javascript/Application/components/Assessment/Item/ItemNaCheckbox.js
KShresthaCwds: Used it from ItemHelper.js

app/javascript/Application/components/Assessment/ReassessmentHelper.js
KShresthaCwds: Done

app/javascript/Application/components/Assessment/ReassessmentHelper.js
KShresthaCwds: Yeah Thank you. Removed It

PR #634, opened 2019-04-09
CANS-837. Supervisor Dashboard: Add "Reassessments Needed" column

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-837

## Description
Adds a column with two pill badges to Assigned Staff table on the Supervisor dashboard

N.B. Comes together with https://github.com/ca-cwds/cans-api/pull/250

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
```diff
@@ -0,0 +1,14 @@
+import moment from 'moment'
+
+const THIRTY_DAYS = 30
+
+export const countWarningsAndDues = reminderDates => {
+  const today = moment()
```
TylerOSI: if this is just a date, i recommending using `moment({hours: 0})` in order to avoid strange edge cases with comparing moments that have different hours.

app/javascript/Application/components/common/DataGridHeader.test.js
```diff
@@ -11,6 +11,15 @@ describe('<DataGridHeader />', () => {
...
 
+  it('has a node title', () => {
+    const wrapper = render(
+      <div>
+        Hello<br />World
+      </div>
+    )
+    expect(wrapper.text()).toContain('HelloWorld')
+  })
+
```
TylerOSI: what does this test? It seems like its just testing the render method, which I assume is doing a shallow render of this div? Maybe its pointless?

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
```diff
@@ -0,0 +1,14 @@
...
+  const todayMinus30Days = moment().subtract(THIRTY_DAYS, 'days')
```
TylerOSI: is this supposed to be 30 days or one month? a month can vary in number of days -- which is why I am asking.

TylerOSI: actually, it looks like the story says 30 days so this is correct.

app/javascript/Application/components/Staff/ReassessmentsNeededCellBadge.js
```diff
@@ -0,0 +1,27 @@
...
+import { Badge } from '@cwds/components'
+
+const ReassessmentsNeededCellBadge = ({ color, number }) => (
+  <div className="reassess-needed-badge">
+    {number ? (
+      <Badge pill color={color}>
+        {number}
+      </Badge>
+    ) : (
+      0
```
TylerOSI: is this zero supposed to be in the badge?

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
```diff
@@ -0,0 +1,45 @@
...
+    it('ignores today and future dates', () => {
+      const warningsAndDues = countWarningsAndDues(['2019-01-01', '2019-01-02'])
+      expect(warningsAndDues.warningsCount).toBe(0)
+    })
+  })
+
+  describe('duesCount', () => {
+    it('ignores due dates that are more than 30 days before today', () => {
+      const warningsAndDues = countWarningsAndDues(['2018-11-29', '2018-12-01'])
+      expect(warningsAndDues.duesCount).toBe(2)
```
wmitchell-cambria: I am a little confused by these tests. If they are ignored, why is the count 2? For the next test, if it is counted, why is the result 0?

app/javascript/Application/components/Staff/ReassessmentsNeededCellBadge.test.js
```diff
@@ -0,0 +1,21 @@
...
+  it('renders plain 0 when number prop is 0', () => {
+    const wrapper = render('warning', 0)
+    expect(wrapper.find('div.reassess-needed-badge').props().children).toEqual(0)
+  })
+
+  it('renders pill badge when number prop is more than 0', () => {
+    const wrapper = render('danger', 1)
+    const badge = wrapper.find('div.reassess-needed-badge').find(Badge)
+    expect(badge.props().pill).toBeTruthy()
+    expect(badge.props().color).toEqual('danger')
```
wmitchell-cambria: What about a test case for the other `color`?

app/javascript/Application/components/Staff/StaffTable.js
```diff
@@ -50,6 +51,25 @@ const columns = [
...
+    id: 'reminderDates',
+    Header: (
+      <DataGridHeader
+        title={
+          <span>
+            Reassessments<br />Needed
+          </span>
+        }
+        tooltip={
+          'Orange indicates that a reassessment is coming due in 30 days or less. Red indicates that a reassessment is past due.'
```
wmitchell-cambria: The colors are provided by the component library. We should not refer to them as 'Orange' and 'Red' here, since the component library could change them. For example, warning could become yellow. We should ask product for a different description.

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
```diff
@@ -0,0 +1,14 @@
...
+  const warningsCount = reminderDates.filter(isoDate => {
+    const reminderDate = moment(isoDate)
+    return reminderDate.isSameOrAfter(todayMinus30Days) && reminderDate.isBefore(today)
```
wmitchell-cambria: This method uses `isSameOrAfter` and 30 days, whereas @TylerOSI has a method using `isAfter` and 1 month. Please coordinate with him to use a single approach. Multiple implementations leads to edge case bugs.

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
```diff
@@ -0,0 +1,14 @@
+import moment from 'moment'
+
+const THIRTY_DAYS = 30
```
wmitchell-cambria: A constant like this is OK, but it does not do a great job of explaining why the value is 30, which is the point of the magic number linter. If we were not allowed to use the number in the name, what would we call it? Perhaps `const REMINDER_PERIOD` or `const WARNING_PERIOD`.

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
AK56: The test description seem typo to me..

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
```diff
@@ -0,0 +1,45 @@
...
+    })
+
+    it('counts in upcoming due dates that are less than 30 days before today and before today', () => {
```
AK56: likewise here

app/javascript/Application/components/Staff/ReassessmentsNeededCellBadge.js
AK56: I think 0 is going to be rendered plain here

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
denys-davydov: Thank you, renamed to `WARNING_PERIOD`

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
denys-davydov: Thank you for the hint, updated to adopt your recommendation

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.js
denys-davydov: Thank you, we are coordinating this issue to have the same behavior

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
denys-davydov: Sorry for the confusion. The tests are reworked after refactoring, I double checked not to have mistakes

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
denys-davydov: Thank you, this test is reworked as well.

app/javascript/Application/components/Staff/ReassessmentsNeededCellBadge.js
denys-davydov: @TylerOSI Yes, @AK56 is right, as per user-story requirements `0` is rendered plan

app/javascript/Application/components/Staff/ReassessmentsNeededCellBadge.test.js
denys-davydov: Actually the `color` prop propagation down to the Badge component is tested here. I don't think we need to cover all the possible values for the propagation. We do not do it usually.

app/javascript/Application/components/Staff/StaffTable.js
denys-davydov: Thank you, Will. Since this is a business requirements matter the discussion is moved to the Jira ticket https://osi-cwds.atlassian.net/browse/CANS-837

app/javascript/Application/components/common/DataGridHeader.test.js
denys-davydov: Thank you, removed the test as a redundant.

TylerOSI:
looking better, I can use the isWarningDate, isPastDueDate, and diffInDays functions in my PR or a later refactoring effort once this is merged. That should alleviate any concerns about duplicated logic.

app/javascript/Application/components/Staff/ReassessmentsDueCalculator.test.js
```diff
@@ -0,0 +1,42 @@
+import { countWarningsAndDues } from './ReassessmentsDueCalculator.js'
+import moment from 'moment'
+
+describe('#countWarningsAndDues()', () => {
+  const dateNowImplTempHolder = Date.now
+
+  beforeAll(() => {
+    const todayMock = moment('2019-01-01')._d // this date will be "today" in tests
```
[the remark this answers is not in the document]
denys-davydov: It will not work with `.day()` but I am surprised that it still works when `._d` is removed. 
Due to a fact that Moment.js creates a wrapper for the Date object and holds it in `_d` field I am surprised it works even without the `._d` part. So I am just removing it now.

denys-davydov:
Retest this please

denys-davydov:
Retest this please

PR #638, opened 2019-04-10
CANS-1065. Improve Retain "number of rows" selected by user per session

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1065

## Description
This will clear session storage on user logout

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Header/UserMenu.js
```diff
@@ -3,14 +3,15 @@ import PropTypes from 'prop-types'
...
+import { clearStorage } from '../../util/sessionStorageUtil'
 
 const logoutPath = logoutUrl()
 
 const UserMenu = ({ user }) => {
   const name = user ? formatUserName(user) : 'Not Available'
   return (
     <UncontrolledUserMenu label={name}>
-      <MenuItem className={'no-uppercase'} tag={'a'} href={logoutPath}>
+      <MenuItem className={'no-uppercase'} tag={'a'} href={logoutPath} onClick={clearStorage}>
```
wmitchell-cambria: There is potentially an edge case here if something prevents the navigation (such as unsaved changes in an assessment). If the user cancels that navigation, then the session store will still be cleared, right? It is unclear to me if this is a critical issue.

denys-davydov: That is a great catch, thank you! I will work on the corner case.

app/javascript/Application/components/common/TimeoutWarning.js
```diff
@@ -36,7 +36,8 @@ export class TimeoutWarning extends Component {
   }
 
   logout() {
-    window.location.href = logoutUrl()
+    SessionStorageUtil.clearStorage()
+    window.location.assign(logoutUrl())
   }
```
DmitriyRudenko: Since session cleaning is needed for logout only, so how about to expose logout service( or util method) which will do all logout related stuff, and do not use sessionStorage directly everywhere?

denys-davydov: This is a JS browser local session storage https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

DmitriyRudenko: Yep, you right it's browser only. It could be better to use it where it's really necessary, in this particular case - logout process, so consider to introduce logout service/utility function

denys-davydov: Not sure I understand you correctly. Please check `UserMenu.js:9`, is it what you mean?

DmitriyRudenko: yes

PR #641, opened 2019-04-12
CANS-831 sort reminder date and left justify the cell content

TylerOSI:
## JIRA
[CANS-831](https://osi-cwds.atlassian.net/browse/CANS-831)

## Description
The reminder date should still be sorted by date, and the cell content also should be left justified.
## Tests

## Types of changes

## Checklist:


TylerOSI:
Jenkins test this please

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplate.js
```diff
@@ -51,8 +52,9 @@ and meet the setting rules of reactTable at same time */
         />
       ),
       headerStyle: { textAlign: 'center' },
+      accessor: client => isoToLocalDate(client.reminder_date),
       Cell: ReminderDateCell,
-      className: 'client-list-table-cell-center',
+      sortMethod: sortDOB,
```
denys-davydov: Not a blocker but it would be nice to rename `sortDOB` function into smth like `sortDates` since having `DOB` in the name is too specific.

TylerOSI:
i re-requested review because i changed sortDOB to sortDate and used moment instead of javascript Date function

TylerOSI:
there still seem to be some issues with sort for reminder date (but not DOB) working through them now

app/javascript/Application/components/Client/ClientSocialWorkerCardTemplateHelper.js
```diff
@@ -1,16 +1,18 @@
...
 export function renderDate(datetime) {
   return !datetime || datetime === null ? null : isoToLocalDate(datetime)
 }
 
-export function sortDOB(a, b) {
-  const dobA = new Date(a).getTime()
-  const dobB = new Date(b).getTime()
-  return dobA > dobB ? 1 : -1
+export function sortDate(a, b) {
+  const dateA = moment(a, LOCAL_DATE_FORMAT)
```
denys-davydov: I think that the performance of the old implementation should be better than the performance of the proposed option. I base this assumption on the fact that the `moment` object is a wrapper around a `Date` object, so it has some additional logic and calculations around the date. But to compare dates we do not actually need those convenient blows and whistles of the moment object, have a unix timestamp is enough for us here, and the `Date` object provides it. The `moment` library is still a great and very helpful thing for us though.

TylerOSI:
simple solution now that works just like it used to. ill merge when the build passes.

PR #642, opened 2019-04-12
CANS-837. Break "Reassessments Needed" column on Supervisor dashboard

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-837

## Description
This will break "Reassessments Needed" column on Supervisor dashboard into two columns: Reassessments Due and Reassessments Past Due

## Tests

## Types of changes

## Checklist:


denys-davydov:
Retest this please

app/javascript/Application/components/Staff/StaffTable.js
```diff
@@ -75,7 +92,7 @@ const columns = [
 const StaffTable = ({ staff }) => (
   <SessionDataGrid
     data={staff}
-    columns={columns}
+    columns={columns(today())}
```
TylerOSI: I dont really consider it necessary at this point to pass today into the staff table columns because its only used in one place. However, I do not consider it a blocking issue to implement it this way.

app/javascript/Application/components/Client/ReminderDateCell.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Badge } from '@cwds/components'
-import { isoToLocalDate } from '../../util/dateHelper'
-import { today, isWarningDate, diffInDays, WARNING_PERIOD } from '../Staff/ReassessmentsDueCalculator'
+import { isoToLocalDate, today } from '../../util/dateHelper'
+import { isWarningDate, diffInDays, WARNING_PERIOD } from '../Staff/ReassessmentsDueCalculator'
```
TylerOSI: are these in use? The rest of the diff for this file does not show `isWarningDate`, `diffInDays` or `WARNING_PERIOD` being used.

denys-davydov: Yes, they are in use. BTW, eslint fails the build when there is an unused import, good linting 👍

app/javascript/Application/components/Staff/StaffTable.js
denys-davydov: The idea here is to use the same today's date for all the reminder_dates on the page, when we are comparing them. There are at least 2 reasons for it: 
1. Consistency. The today date is set only once for the table, so all the table logic is using the same value (consider the date transition moment of 12:00am when some records are before it and are using the previous day while others are using the next day for some logic and calculations).
2. Performance. The Date object is initialized only once instead of multiple times for each reminder_date.

app/javascript/Application/components/Client/ReminderDateCell.js
TylerOSI: on my bad i miss-read the diff. after the changes its importing the exact same thing

app/javascript/Application/components/Staff/StaffTable.js
TylerOSI: right that makes sense. I didn't realize that todaysDate was used in two places. so 👍

denys-davydov:
retest this please

PR #646, opened 2019-04-15
CANS-1079 - Expand All / Collapse All edge cases

wmitchell-cambria:
## JIRA

[CANS-1079 - Assessment form - EXPAND ALL / COLLAPSE ALL functionality](https://osi-cwds.atlassian.net/browse/CANS-1079)

## Description

This adds some functionality to the `Expand All / Collapse All` buttons to be a little smarter about what state the domains are currently in. Specifically, it handles the cases where the user:
1. Manually expands all domains (The button will now switch to Collapse All if it was Expand All)
2. Manually collapse all domains (The button will not switch to Expand All if it was Collapse All)

In order to support this functionality, I lifted the domain `isExpanded` state to be tracked in one location for all domains (so that we can check if they are all the same). This is isolated to a `DomainExpansionController` that sits in the middle of the Assessment component (now split between Assessment for update functionality and AssessmentCard for render functionality).

Along the way, I was able to make some performance improvements. When expanding and collapsing domains, previously all domains and items in them would re-render, making it very slow. Now expanding and collapsing only affects the domain targeted, and even Expand All / Collapse All does a lot less re-rendering than it did before. Try it. It's quite fast. 🚀 

I had fun with this one.

## Tests

I feel that the Expand All / Collapse All functionality is isolated well enough to the `DomainExpansionController` that a Selenium test for it would be redundant. (Also, this is not necessarily _critical_ functionality if it fails.) Existing tests should pass.

## Types of changes

## Checklist:

- [x] I ran all my tests locally which were green. _(I didn't run acceptance tests yet._)


app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -142,86 +140,82 @@ class Assessment extends Component {
...
+      onConfidentialityUpdate: this.handleUpdateItemConfidentiality,
+      onAddCaregiverDomain: this.addCaregiverDomainAfter,
+      onRemoveCaregiverDomain: this.removeCaregiverDomain,
+      onCaregiverNameUpdate: this.updateCaregiverName,
+      onDomainCommentUpdate: this.updateDomainComment,
+      onDomainReviewed: this.updateDomainIsReviewed,
+      handleWarningShow: this.props.handleWarningShow,
+    }
+
+    return !(isUnderSix === null || isUnderSix === undefined) ? (
```
TylerOSI: this should be able to be rewritten as:
```
isUnderSix ? (...) : ...
```

app/javascript/Application/components/Assessment/Assessment.test.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react'
-import { Assessment, Domain } from './'
 import { shallow } from 'enzyme'
+import { Assessment } from './'
```
TylerOSI: you can omit the filename?

app/javascript/Application/components/Assessment/AssessmentCard.js
```diff
@@ -0,0 +1,107 @@
...
+      expandCollapse={onExpandCollapseAll}
+    />
+    <CardBody>
+      {domainsExpanded.map(({ domain, isExpanded }, index) => {
+        const { id, code, caregiver_index: caregiverIndex } = domain
+        const domainI18n = getI18nByCode(i18n, code)
+        return (
+          <Domain
+            index={index}
+            key={code + caregiverIndex + id}
```
TylerOSI: guaranteed uniqueness?

app/javascript/Application/components/Assessment/AssessmentCard.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+  it('renders a card with header, body, and footer', () => {
+    const card = render().find(Card)
+    expect(card.exists()).toBe(true)
+    expect(card.find(DomainsHeader).exists()).toBe(true)
+    expect(card.find(CardBody).exists()).toBe(true)
+    expect(card.find(CardFooter).exists()).toBe(true)
+  })
+
+  it('renders footer', () => {
```
TylerOSI: why test that it renders the footer, when it clearly contains a footer in the above test?

app/javascript/Application/components/Assessment/AssessmentContainerInner.js
```diff
@@ -109,17 +102,16 @@ class AssessmentContainerInner extends Component {
           assessmentStatus={assessment.status}
           domains={assessment && assessment.state && assessment.state.domains}
           i18n={i18n}
-          isUnderSix={Boolean(assessment && assessment.state && assessment.state.under_six)}
+          isUnderSix={isUnderSix}
           disabled={!isEditable}
         />
         <Assessment
+          key={isUnderSix.toString()}
```
TylerOSI: if this is a boolean, the key is one of two values "true" or "false". Is there any risk of duplicated keys here?

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -125,88 +122,82 @@ class Domain extends Component {
...
-                  disabled={this.props.disabled}
-                  domainBottomCollapseClick={this.handleExpandedChange}
+            <Grid item xs={8}>
+              <Typography variant="title" style={{ color: '#0e6f89', fontSize: 16 }}>
+                <Icon
+                  id={`domain${index}-expand`}
+                  className="domain-icon"
+                  icon="chevron-down"
+                  size="lg"
+                  rotation={isExpanded ? null : ROTATION_RIGHT}
```
TylerOSI: how does this behave with null rotation?

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -125,88 +122,82 @@ class Domain extends Component {
...
-          )}
-        </ExpansionPanel>
-      </Fragment>
+                <span className="domain-item-margin">{title}</span>
+                <Icon className="domain-help-icon" icon="info-circle" id={`domain-${index}`} />
+                {description ? (
+                  <UncontrolledTooltip style={{ minWidth: '20rem' }} target={`domain-${index}`} placement="top">
+                    {description}
+                  </UncontrolledTooltip>
+                ) : null}{' '}
```
TylerOSI: does this need to be a non-breaking space -- since normal spaces are ignored in HTML elements (except for text)? If it is a non-breaking space be sure to code it as `&nbsp;`

app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -125,88 +122,82 @@ class Domain extends Component {
...
+                {(isCaregiverDomain && caregiverName === '') ||
```
TylerOSI: why do you need this statement given the statement below?
```
(isCaregiverDomain && caregiverName && caregiverName.trim() === '')
```
since `"".trim()` is an empty string?

app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -41,6 +42,7 @@ const domainComponentDefault = (
     index={1}
     onItemCommentUpdate={() => {}}
     onDomainCommentUpdate={() => {}}
+    onExpandedChange={() => {}}
```
TylerOSI: empty function everywhere!!!

app/javascript/Application/components/Assessment/Assessment.js
wmitchell-cambria: It can't, actually. `isUnderSix` has three states (true/false/undefined). When it is undefined, the user has not yet selected an Age Template, and the lower portion of the assessment is not rendered. Once it is selected (true _or_ false), this renders.

app/javascript/Application/components/Assessment/Assessment.test.js
wmitchell-cambria: Yeah...without a filename, it imports the `index.js` file in that directory. Assessment happens to be re-exported from that file, since it is used in other directories. It's not my favorite pattern.

app/javascript/Application/components/Assessment/AssessmentCard.js
wmitchell-cambria: Yup, should be. The key previously used these as well as two additional (and clearly not unique) pieces:
```key={code + caregiverIndex + isDefaultExpanded + id + isUnderSix}```

It moved files when it changed, which made it harder to diff, sorry.

`isDefaultExpanded` is no longer necessary since the Expand All / Collapse All are managed through the controller component. (Previously, we would change the default and force all components to remount to reset the state.)
Similarly, `isUnderSix` is not needed here, as I moved it to the Assessment component. When you switch age templates, I felt it made more sense for the Assessment to remount, rather than individual domains. This also has the desired side effect of resetting the DomainExpansionController when the template changes, which is what the key was used for before.

Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/AssessmentCard.js]: Err, additional notes on the fields themselves, for knowledge sharing:
* `code` is not unique alone because there can be multiple caregiver domains with the same code.
* `code + caregiver_index` is unique but not desirable as a key. If you add a caregiver domain in the middle of your list, they will all shift caregiver_index (a, b => a, b, c). The `id` is used to track the domains as they reorder (1, 2 => 1, 3, 2).

File: app/javascript/Application/components/Assessment/AssessmentCard.test.js (hunk 4 above)
Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/AssessmentCard.test.js]: Two reasons:
1. I wanted to copy as many tests from `Assessment.test.js` as possible, to make sure I wasn't losing functionality.
2. This does test it a little differently. The above test ensures that there is a footer, and this test ensures that the footer contains the thing you tell it to contain.

File: app/javascript/Application/components/Assessment/AssessmentContainerInner.js (hunk 5 above)
Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/AssessmentContainerInner.js]: Keys only have to be unique among the children of an element, in this case the Fragment. It may be slightly safer to say `\`assessment-${isUnderSix}\``, in case other children get keys. React will throw a loud error if there are accidental duplicates, though, so the risk is pretty low.

File: app/javascript/Application/components/Assessment/Domain.js (hunk 6 above)
Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/Domain.js]: If I remember correctly, the Icon restricts rotation to be `90`, `180`, or `270`. Instead of saying `rotation={0}`, we set it to null, which tells it not to rotate at all. (It's possible this has changed... note that this was in the previous version as well, a few lines down. This file was mostly indentation changes.)

File: app/javascript/Application/components/Assessment/Domain.js (hunk 7 above)
Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/Domain.js]: For this and the next...I have no idea. They were like this before. :-( For `Domain.js` in particular, it may be easier to see the 'real' changes if you set GitHub's Diff settings at the top of the page to `Hide whitespace changes`.

File: app/javascript/Application/components/Assessment/Domain.test.js (hunk 9 above)
Comment (wmitchell-cambria) [reply app/javascript/Application/components/Assessment/Domain.test.js]: Yeah, some of these test files are not super well abstracted. I may take a second pass at cleaning this up in a new PR if I have time. Adding a `render` function would go a long way.

**Comment by wmitchell-cambria** [comment]
I'm working on the acceptance test failure. Because I am filtering out the non-rendered Domains higher up in the three, the generated `ids` for the HTML elements are different.

File: app/javascript/Application/components/Assessment/AssessmentCard.js (hunk 3 above)
Comment (abelayt) [reply app/javascript/Application/components/Assessment/AssessmentCard.js]: Isn't 'isDefaultExpanded' just renamed 'isUnifiedExpansion' wherever it was used? I see 'isDefaultExpanded' is removed from Assessment.js and 'isUnifiedExpansion' is used on the new AssessmentCard. Also 'isDefaultExpanded' is still being used in Assessment.test.js and needs to be removed.

File: app/javascript/Application/components/Assessment/Assessment.js [hunk 10]
```diff
@@ -142,86 +140,82 @@ class Assessment extends Component {
...
+            return { ...domain, is_reviewed: true }
+          }
+          return domain
+        }),
+      },
+    }
+
+    if (isChanged) {
+      this.props.onAssessmentUpdate(newAssessment)
+    }
```
AK56: +1 for the performance tweak here

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -142,86 +140,82 @@ class Assessment extends Component {
...
+          footer={footer}
+          i18n={i18n}
+          isCompletedAssessment={isCompletedAssessment}
+          isDomainsReviewed={isDomainsReviewed}
+          isUnderSix={isUnderSix}
+          isUsingPriorRatings={Boolean(assessment.preceding_assessment_id)}
+          previousRatingsMap={previousRatingsMap}
+        />
+      </DomainExpansionController>
+    ) : null
```
AK56: 💯 on refactor and `AssessmentCard` tie up

app/javascript/Application/components/Assessment/DomainExpansionController.js
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+class DomainExpansionController extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      domains: [],
+      domainsExpanded: [],
```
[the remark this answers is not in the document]
wmitchell-cambria: I agree. The issue, though, is that `id` is marked as optional in our data model, so I don't want to fully rely on it. As it is now, it works even if we change the unique key of the domain. With only a handful of domains, iterating over the array to find things shouldn't be too much of an issue. The much greater performance issue is re-render cycles.

One other side advantage of storing this as an array is that the expansion state is stored with the domain in order, so children can iterate over a single array rather than iterating over domains and then checking the map for an isExpanded value.

app/javascript/Application/components/Assessment/Assessment.js
```diff
@@ -142,86 +140,82 @@ class Assessment extends Component {
...
+          if (domain.caregiver_index === caregiverIndex && domain.code === code && domain.is_reviewed !== true) {
+            isChanged = true
+            return { ...domain, is_reviewed: true }
+          }
+          return domain
+        }),
+      },
+    }
+
+    if (isChanged) {
```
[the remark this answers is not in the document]
wmitchell-cambria: Good point. In addition to saving the iteration, it will also allow me to avoid the side-effect style of `isChanged` here.

app/javascript/Application/components/Assessment/DomainExpansionController.js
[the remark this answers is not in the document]
denys-davydov: Got it, that is fine then.

app/javascript/Application/components/Assessment/Assessment.js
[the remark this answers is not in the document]
denys-davydov: Comment after the change: the `Assessment#updateDomainIsReviewed()` can be reverted now to the previous state.

app/javascript/Application/components/Assessment/AssessmentCard.js
wmitchell-cambria: Good catch on `Assessment.test.js`. That was dead code. I just removed it.

In most places, I removed `isDefaultExpanded`, because the `isExpanded` state was tracked individually with no default. Then I added `isUnifiedExpansion` to track what the last unified state was, and only just now realized that it could probably have still been called `isDefaultExpanded`. I can change it if you think `isUnifiedExpansion` is confusing. It isn't my favorite name.

abelayt: No need to change it since isDefaultExpanded is confusing too. But let's just keep in mind these two, defaultExpanded(isDefaultExpanded) & expanded (isExpanded) are coming from Material UI ExpansionPanel API. I think we need to make sure they mean/used the same in our app too.

wmitchell-cambria: Oh, interesting. I didn't realize that Material had a defaultExpanded as well. Good to know.

wmitchell-cambria:
Retest this please, Jenkins. I know that span exists, I can see it in the screenshot...

denys-davydov:
Retest this please

denys-davydov:
Retest this please

denys-davydov:
One regression scenario is stable in failing, tried it locally - and it is fail again. So it is not an environment issue. 
regression_failed

denys-davydov:
Oh! But the second time it passed locally.

AK56:
Yes, it passed for me too locally, so Jenkins Retest this please

denys-davydov:
Retest this please

denys-davydov:
Retest this please

denys-davydov:
Retest this please

denys-davydov:
Retest this please

PR #648, opened 2019-04-15
CANS-831 refactor date sorting and add sort method to reminder date col

TylerOSI:
## Description
The default sort mechanism for reminder date does not work anymore. It used to only sort dates within one month of today (that is why default sort worked). I needed to use a sortMethod. In tandem, I refactored date sorting into one method called `sortDate` and removed `sortDOB`

## Tests

## Types of changes

## Checklist:


app/javascript/Application/util/dateHelper.test.js
```diff
@@ -23,7 +24,7 @@ describe('#isoToLocalDate()', () => {
   })
 
   it('returns Invalid Date when input is undefined', () => {
-    expect(isoToLocalDate(undefined)).toEqual('Invalid date')
+    expect(isoToLocalDate(undefined)).toEqual(undefined)
```
wmitchell-cambria: Just curious:  What was the reasoning behind this change?

We should update the test name, as well.

app/javascript/Application/util/dateHelper.js
```diff
@@ -135,3 +135,12 @@ export function calculateDateDifferenceInYears(startDate, endDate) {
   const duration = moment.duration(moment(endDate).diff(moment(startDate)))
   return Math.floor(duration.asYears())
 }
+
+export function sortDate(a, b) {
+  if (a && b) {
+    return moment(a, LOCAL_DATE_FORMAT) - moment(b, LOCAL_DATE_FORMAT)
+  }
+  if (a) return 1
```
wmitchell-cambria: These three lines (143-145) are not covered in unit tests. Since it's important enough to have logic for null and non-date inputs, let's test them. I'd be interested in falsy values as well as truthy values that do not parse as dates.

app/javascript/Application/util/dateHelper.test.js
TylerOSI: the function `isoToLocalDate` used to return null when passed a null date, and a string 'Invalid date' when passed undefined. It now returns null and undefined respectively.

app/javascript/Application/util/dateHelper.js
TylerOSI: good point ill update a test or add a test case

app/javascript/Application/util/dateHelper.test.js
TylerOSI: also `moment(undefined)` is now, so its dangerous to pass anything falsy into moment)

TylerOSI:
there was a linting issue. waiting for the PR build to pass before merging.

PR #666, opened 2019-04-19
CANS-1157 - Refactor caregiver indexing to avoid UNSAFE_componentWillReceiveProps

wmitchell-cambria:
## JIRA
[CANS-1157 - Replace Unsafe Components](https://osi-cwds.atlassian.net/browse/CANS-1157) (Partial)

## Description

This refactors how caregiver indexes ('a', 'b', 'c', etc) are tracked within our component tree. By moving them up from Assessment to AssessmentContainer's `updateAssessment` function, we can ensure that any update results in the caregiver indexes being in consecutive alphabetical order. To support this, some of our tests switch to using domain `id`, and the actual indexing method is moved out to an `AssessmentMutations` helper (which could also start including things like `updateItem`, etc).

Anyways, this refactor allows Assessment to avoid calling back with an initial update of the caregiver indexes on componentWillReceiveProps, since the AssessmentContainer is now responsible for tracking those indexes on its own.

The indexing method was also rewritten so that it does not modify any objects in place, instead returning new objects that React can diff more efficiently.

~(I am creating this as a draft PR for now, as acceptance test results are still very unknown.)~ Tests look good, and I confirmed slightly different indexing behavior with Sonoltra.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentMutations.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+
+      newDomains.push(newDomain)
+      return { newDomains, numChanged: newNumChanged, nextCaregiverIndex: newNextCaregiverIndex }
+    },
+    { newDomains: [], numChanged: 0, nextCaregiverIndex: 0 }
+  )
+
+  const { newDomains, numChanged } = result
+  return numChanged > 0 ? newDomains : domains
```
denys-davydov: Not sure if this will improve the performance since the domains we are working here with are not the original domains from the components state, but their clones, so eventually the component will still receive new objects to the state.

wmitchell-cambria: I think that's true in some cases, yes. I made sure that the function itself doesn't clone the domains unless something changes. If we continue to use that pattern elsewhere, we should see much better performance in terms of React rendering. Excessive cloning is a big cause of that right now.

My idea going forward would be that this Mutations file would grow to contain all the ways that we can update an Assessment, and that those updates would all behave in the same way (returning copies that only clone the pieces that changed). That would help isolate the behavior outside of our more complex components will providing better performance overall. At that point, a library like Immutable.js (used by Intake) could potentially help as well.

denys-davydov: That is a good idea and a type of "default" or "normal" approach to component state management, while we've been using a kind of "short cut" approach that helped us to deliver significantly faster by paying with a performance. Looks like we are getting out of the spare performance resources and we probably need to undertake some actions now. Applying this approach will dramatically increase the complexity of the assessment mutation functions (and this `updateCaregiverDomainsIndices()` function implementation is a great example of the assertion). Due to that the refactoring of the critical code could be bug-capacious and becomes highly risky in the light of the upcoming CANS 2.0 release.

PR #667, opened 2019-04-19
Cans 1126 fix assessment history bug

TylerOSI:
## JIRA
[CANS-1126](https://osi-cwds.atlassian.net/browse/CANS-1126)

## Description
There are two places where components are rendered to early with misleading results. This is an effort to fix both of those problems.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentContainerInner.js
```diff
@@ -102,11 +107,11 @@ class AssessmentContainerInner extends Component {
           assessmentStatus={assessment.status}
           domains={assessment && assessment.state && assessment.state.domains}
           i18n={i18n}
-          isUnderSix={isUnderSix}
+          isUnderSix={this.isUnderSix(assessment)}
```
denys-davydov: So now the `isUnderSix()` function will be called twice: here and in the line 114; while previously it was a variable that was initialized once and the value was used twice. Not a huge deal in this particular case, but we probably need a reason to reduce performance.

app/javascript/Application/components/Assessment/AssessmentFormHeader.js
```diff
@@ -233,6 +238,7 @@ AssessmentFormHeader.defaultProps = {
 }
 AssessmentFormHeader.propTypes = {
   assessment: PropTypes.object.isRequired,
+  assessmentServiceStatus: PropTypes.string.isRequired,
```
AK56: We can enforce PropTypes to be one of the LoadingState values:
`PropTypes.oneOf(Object.values(LoadingState)).isRequired`

app/javascript/Application/components/Client/ClientAssessmentStatisticsHeader.js
```diff
@@ -3,22 +3,25 @@ import PropTypes from 'prop-types'
...
-        <AssessmentRecordControl {...assessmentRecordControlProps} />
+        {ready && <AssessmentRecordControl {...assessmentRecordControlProps} />}
       </CardTitle>
     </CardHeader>
   )
 }
 
 ClientAssessmentStatisticsHeader.propTypes = {
   isComparisonShown: PropTypes.bool.isRequired,
+  loadingState: PropTypes.string.isRequired,
```
AK56: Same here, enforce the PropTypes to be one of the LoadingState value

app/javascript/Application/components/Assessment/AssessmentContainerInner.js
TylerOSI: with another approach more refactoring would need to be done for this simple change. only reason i did this was to make code climate happy.

app/javascript/Application/components/Assessment/AssessmentFormHeader.js
TylerOSI: good point. we can do that. Setting proptypes does not enforce the values though; It throws an error in the console so that you can "catch" the problem. If we really wanted to enforce types, we would need to use Flow or TypeScript.

AK56: Ahh, good to know :)

PR #674, opened 2019-04-23
CANS 1179 headless chrome

TylerOSI:
## JIRA
[CANS-1179](https://osi-cwds.atlassian.net/browse/CANS-1179)
## Description
Use headless chrome instead of Xvfb. It may increase performance of the tests.

## Tests

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,9 @@
     "test:rspec": "bundle exec rspec --exclude-pattern 'spec/{acceptance,a11y,regression}/**/*_spec.rb' --format NyanCatFormatter",
     "test:acceptance": "bundle exec rspec spec/acceptance",
     "test:regression": "bundle exec rspec spec/regression --format html --out regression-report/index.html",
-    "test:a11y": "bundle exec rspec spec/a11y"
+    "test:regression:local": "HEADLESS='false' bundle exec rspec spec/regression",
+    "test:a11y": "bundle exec rspec spec/a11y",
+    "test:a11y:local": "HEADLESS='false' bundle exec rspec spec/a11y"
   },
   "dependencies": {
     "@babel/core": "^7.0.0",
--- a/spec/acceptance_helpers/login_helper.rb
+++ b/spec/acceptance_helpers/login_helper.rb
@@ -9,11 +9,8 @@ module LoginHelper
   end
 
   def enter_credentials(login_config = default_json)
-    js_script = 'arguments[0].focus();' \
-                 'arguments[0].setAttribute("value", arguments[1]);' \
-                 'arguments[0].blur();'
-    input_field = find('input#username')
-    page.execute_script(js_script, input_field.native, JSON.generate(login_config))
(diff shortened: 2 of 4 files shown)


TylerOSI:
jenkins please test this

TylerOSI:
Jenkins test this please

package.json
```diff
@@ -14,7 +14,9 @@
     "test:rspec": "bundle exec rspec --exclude-pattern 'spec/{acceptance,a11y,regression}/**/*_spec.rb' --format NyanCatFormatter",
     "test:acceptance": "bundle exec rspec spec/acceptance",
     "test:regression": "bundle exec rspec spec/regression --format html --out regression-report/index.html",
-    "test:a11y": "bundle exec rspec spec/a11y"
+    "test:regression:local": "HEADLESS='false' bundle exec rspec spec/regression",
```
AK56: we should add the `--out` flag for the results output file when running locally as well. Also if we are utilizing an `HEADLESS` env var we might not need that option in the script here

TylerOSI: we need HEADLESS In the script here, because environment variables are not propagated to the running command.

running with or without `--out` is more or less a preference thing. I usually prefer that a report is generated for jenkins, but locally the command is run outputting to the console for easier debugging.

We can sidebar about this in standup today -- decide how we want to do it as a team. If someone wants to run rspec a different way than this script, it can be done on the command line easily enough.

AK56: Is the `HEADLESS=true` necessary in the CL since you are utilizing the `HEADLESS` env var in the acceptance_helper.rb below to set the chrome driver to be headless or not ?

TylerOSI: `HEADLESS` Is true by default in the code bellow, so setting it to false here overrides that.

TylerOSI:
i can also remove xvfb from the testing base image, but that opened up a whole can of worms, so I am going to do a separate PR for that if I can get it done.

spec/acceptance_helpers/login_helper.rb
```diff
@@ -9,11 +9,8 @@ def login(login_config = default_json)
   end
 
   def enter_credentials(login_config = default_json)
-    js_script = 'arguments[0].focus();' \
-                 'arguments[0].setAttribute("value", arguments[1]);' \
-                 'arguments[0].blur();'
-    input_field = find('input#username')
-    page.execute_script(js_script, input_field.native, JSON.generate(login_config))
+    page.fill_in('username', with: JSON.generate(login_config))
```
denys-davydov: Please leave the previous approach with JS script here since it gives a better performance, we are saving 5 to 10 seconds on every login by using it.

package.json
denys-davydov: I would prefer to have an html report for local regression run, it is a lot more verbose hence helpful for debugging/analyzing issues (and it has a screenshot 👍 )

TylerOSI:
comments noted. closing this in favor of a more complete effort which will hopefully be in a new PR

PR #675, opened 2019-04-24
Update Items for better React performance

wmitchell-cambria:
## JIRA
Related to [CANS-1167 - CANS application slow response time in IEv11 on Assessment Page.](https://osi-cwds.atlassian.net/browse/CANS-1167)

## Description

This improves the performance of setting item ratings (as well as other item updates, such as comments), through two refactors:
1. `updateItem` is refactored to return a new Assessment object that shares as many references as possible with the original. This helps React figure out when pieces of the assessment have changed (and more importantly, when they have _not_ changed). With this refactor, setting an item rating does not cause other Domains to re-render. _Because this updateItem method is a bit more complex than it was before, I factored it out into the AssessmentMutations file and broke it into a couple helper pieces that can be reused for other operations._
2. Even with the above refactor, the entire domain and all items within it were re-rendering each time and item was changed. I tracked this down to the i18n object, which was creating a new object for each item render cycle when it filtered the i18n object by item code. Instead, we can just pass the whole object to the item and have it fetch the keys based on code directly.

With these two changes, only the item modified _and no other_ is changed, allowing the UI to react much more efficiently as the user selects ratings.

## Tests

I manually tested this in IE11 in a virtual machine. There was a noticeable performance improvement.

## Types of changes

## Checklist:

- [ ] I ran all my tests locally which were green. _(I have not run acceptance tests yet.)_


app/javascript/Application/components/Assessment/AssessmentMutations.js
```diff
@@ -35,3 +35,50 @@ export const updateCaregiverDomainsIndices = domains => {
   const { newDomains, numChanged } = result
   return numChanged > 0 ? newDomains : domains
 }
+
+const updateIfChanged = (obj, key, value) => {
+  if (obj[key] === value) {
+    return obj
+  }
```
TylerOSI: you may not need this if statement since `return { ...obj, [key]: value }` will return an object that is the same as obj. If the object needs to be the exact same object, then you will need to keep this if statement.

wmitchell-cambria:
Yeah, I was thinking the same thing re: i18n. Ideally, I think we should process it early, once. The approach with this PR is not as bad as it seems, though, since the object is not copied. It is just the single large object passed from place to place. My greatest concern with it would be if the i18n values were changing frequently, which could cause rendering all down the app (which...we would have had happen before as well).

I agree on the complexity of these mutation functions. They are not nearly as clear as I would like. I think there are some ways to improve them, but I'm not sure just yet. Right now, the performance gain is definitely worth it, and I tried to add significant test coverage to make up for the added complexity.

app/javascript/Application/components/Assessment/AssessmentMutations.js
wmitchell-cambria: It does need to be the exact same object. React does strict equality on its props as they change, and therefore the goal is for them to remain strictly equal unless they change.

PR #676, opened 2019-04-24
Loosen restrictions on references

wmitchell-cambria:
## JIRA

No story. Leaving this as a Draft at the moment, for discussion.

## Description

This is based on some feedback from recent PRs, to try to simplify the code in AssessmentMutations. By loosening some of the requirements that references must be the same when no value changed, we can remove some additional checks and state tracking. However, this does have performance implications. Even if every item is the same, the list of items will be different, causing DomainItemList to re-render (for each Domain). This is not as bad as rendering every item, but may still be noticeable.

This PR basically subjects us to the issues that Immutable.js suffers from, where mapping without changing any values still returns a new Immutable object.

## Tests

## Types of changes

## Checklist:

### DIFF
--- a/app/javascript/Application/components/Assessment/AssessmentMutations.js
+++ b/app/javascript/Application/components/Assessment/AssessmentMutations.js
@@ -2,38 +2,30 @@ const INDICES = 'abcdefghijklmnopqrstuvwxyz'
 
 export const updateCaregiverDomainsIndices = domains => {
   const result = domains.reduce(
-    ({ newDomains, numChanged, nextCaregiverIndex }, domain) => {
+    ({ newDomains, nextCaregiverIndex }, domain) => {
       let newDomain
-      let newNumChanged
       let newNextCaregiverIndex
 
       const nextCaregiverChar = INDICES.charAt(nextCaregiverIndex)
 
       if (domain.is_caregiver_domain === false) {
         newDomain = domain
-        newNumChanged = numChanged
         newNextCaregiverIndex = nextCaregiverIndex
-      } else if (domain.caregiver_index === nextCaregiverChar) {
-        newDomain = domain
-        newNumChanged = numChanged
-        newNextCaregiverIndex = nextCaregiverIndex + 1
       } else {
         newDomain = {
           ...domain,
           caregiver_index: nextCaregiverChar,
         }
-        newNumChanged = numChanged + 1
         newNextCaregiverIndex = nextCaregiverIndex + 1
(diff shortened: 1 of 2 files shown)


wmitchell-cambria:
Jenkins needs a label. Retest this please.

wmitchell-cambria:
So, one thing to note is that apparently the domains were already re-rendering, due to the same i18n issue as the items. If that is fixed, then this could show a bigger difference.

That is, every time the Assessment renders, Domains see an unnecessary change in their i18n prop. So no matter what we do with the other props, they are updating anyways.

Let me fix that then test with or without this change.

wmitchell-cambria:
That related PR has been created: #677

wmitchell-cambria:
I rebased this on top of the Domain i18n fixes. I think the difference is much more noticeable now.

wmitchell-cambria:
I am leaning away from merging this. I think there is a way to clean up the mutations while maintaining performance benefits, but it would involve a larger refactor of the way these updates are happening. Specifically:

* In `Item.js`, instead of using separate callbacks for each property (`onCommentUpdate`, `onRatingUpdate`, etc), use a single callback for `onItemUpdate`. That callback can set the property on the new item and pass the entire item object back up the hierarchy.
* Higher up, in `Domain.js` for example, we would then do the same thing. Instead of separate callbacks for items and DomainComment, etc, have a single `onDomainUpdate` callback. When the item changes, it can be inserted into the existing domain list while preserving other items.
* At the assessment level, same thing. Take the new domain and set it in the list of domains.

Each of these steps is much simpler, since they are only doing one thing (updating an object, or changing a single item in a list). That way, we also avoid having to iterate over the _entire_ assessment to find a single Item that changed (because our React tree already knows what item changed, what domain it was in, etc).

The effect would be the same, but the helper functions would mostly be unnecessary. It's a lot of code to move around right now, though, so I'm not sure it's worth taking on.

wmitchell-cambria:
I'm going to close this. Because the complex code is isolated to a single file, I don't think it is worth the performance trade-off, given how many performance-related issues we have seen in the past sprint. The alternative refactor described in my previous comment would be a better way to clean up the code while maintaining performance... it just touches too much core functionality for no user-facing benefit to undertake it right before our 'code complete' date.

TylerOSI:
As you mentioned the performance / code quality trade-off here doesn't seem like its worth it. We want to go in a positive direction on both fronts, so I am in support of this decision.

PR #679, opened 2019-04-25
Performance Refactors for Assessment page load and Expand/Collapse

wmitchell-cambria:
## JIRA

No story. (Loosely related to CANS-1167)

## Description

This PR contains three performance-based refactors, all with a shared goal. Currently, there is a tradeoff between the time it takes for an initial assessment to load and render, and the time it takes to Expand domains. This is because Domains, Items, Ratings, etc, are very time consuming to render, especially on IE11. (To some extent, Material UI may be at fault here. It is also just a lot of DOM nodes, though.) The tradeoff is currently set to skip rendering of Items completely when they are hidden. This speeds up initial load, but it slows down Expand All (and other Expand operations). I wanted to find a way for both to be fast. Here is what this commit does, to that end:

1. It extracts the "ExpansionPanelSummary" piece of the Domain out into its own PureComponent. This saves some unnecessary rendering in a lot of cases, such as when the enable/disable state of the assessment changes.
2. Stops tracking `isEditable` in the AssessmentContainer state. This is entirely derived from the Assessment object and its metadata. Tracking it in state resulted in an extra (very expensive) render cycle.
3. In Domain.js, delay the rendering of the DomainItemList by a single render cycle. That way, the domains can render onto the page quickly, and then behind the scenes the longer render can happen. This allows the page to look complete sooner, while ensuring that the Items are there for when the Domains are all expanded later. (Note:  This only takes effect if the domain starts out as collapsed. If it is expanded, we render the whole thing immediately.)

Overall, (3) adds a little complexity to the Domain component. I think the performance trade-off is well worth it. Once we get off of Material UI and onto Component Library, I suspect that rendering the full assessment will be much more efficient.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Domain.js
```diff
@@ -39,6 +39,16 @@ class Domain extends React.PureComponent {
     return null
   }
 
+  componentDidUpdate = async () => {
```
denys-davydov: Not fond of this one. If I was choosing one of three fixes to remove, this would be gone :)

wmitchell-cambria: I definitely agree. I was trying to get rid of the async/setTimeout piece, but it didn't seem to have the desired effect. I would not be happy to see this pattern repeated, but I don't know of another way to get both initial load time _and_ Expand/Expand All to be fast.

PR #680, opened 2019-04-25
CANS-1030: Enhance the "Assessment Conducted By" field

DmitriyRudenko:
## JIRA

https://osi-cwds.atlassian.net/browse/CANS-1030

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -0,0 +1,78 @@
...
+import { Row, Col, Label } from 'reactstrap'
+import PropTypes from 'prop-types'
+import { clone } from '../../../util/common'
+import './style.sass'
+import ConductedByNameField from './ConductedByNameField'
+import ConductedByRole from './ConductedByRole'
+
+class ConductedBy extends Component {
+  handleConductedByFirstNameChange = event => {
+    const assessment = clone(this.props.assessment)
```
wmitchell-cambria: Please avoid using `clone`. It has poor performance impacts, since it copies fields that are not changing. Instead, you can shallowly clone the assessment:

```const assessment = { ...this.props.assessment }```

This will allow the Conducted By fields to be updated without rendering each domain and item.

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js [hunk 2]
```diff
@@ -0,0 +1,78 @@
...
+            />
+          </Col>
+        </Row>
+      </div>
+    )
+  }
+}
+
+ConductedBy.propTypes = {
+  assessment: PropTypes.object.isRequired,
```
wmitchell-cambria: By passing in the entire assessment here, these components will re-render each time the assessment is changed in any way, even if the Conducted By fields are not affected. This can have performance impacts on unrelated user actions, like selecting a rating on an item. Instead, one option would be to pass in the individual fields explicitly as their own props. The event handlers will have to be updated accordingly.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js
```diff
@@ -0,0 +1,99 @@
...
+    const lastNameInput = wrapper.find(ConductedByNameField).at(1)
+    expect(lastNameInput.props().value).toBe(assessment.conducted_by_last_name)
+    expect(lastNameInput.props().disabled).toBe(props.disabled)
+    expect(lastNameInput.props().id).toBe('conducted-by-last-name')
+    expect(lastNameInput.props().placeholder).toBe('Last Name')
+    expect(lastNameInput.props().onChange).toBe(wrapper.instance().handleConductedByLastNameChange)
+    expect(lastNameInput.props().errorMessage).toBe('')
+  })
+
+  it('renders role nput', () => {
```
wmitchell-cambria: typos in these test descriptions:  nput -> input

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -0,0 +1,78 @@
...
+    this.props.onAssessmentUpdate(assessment)
+  }
+
+  render() {
+    const { assessment, disabled } = this.props
+    return (
+      <div>
+        <Row>
+          <Col sm={4}>
+            <Label for={'conducted-by-first-name'} className={'assessment-form-header-label'}>
```
wmitchell-cambria: Because there are more than one input in this group, it is more accurate for this to be the `legend` of a `fieldset`. Individual fields should have their own labels, either as elements or aria attributes.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -0,0 +1,78 @@
...
+              Assessment Conducted By *
+            </Label>
+          </Col>
+        </Row>
+        <Row className={'assessment-form-header-inputs'}>
+          <Col sm={4}>
+            <ConductedByNameField
+              id={'conducted-by-first-name'}
+              placeholder={'First Name'}
```
wmitchell-cambria: [Placeholders should not be used as labels.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Labels_and_placeholders) (I guess this is more of a design concern than a development concern, but I do need to continue to point it out.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js
```diff
@@ -0,0 +1,99 @@
...
+    jest.resetAllMocks()
+    const event = {
+      target: {
+        value: `${assessment.conducted_by_last_name}new`,
+      },
+    }
+    const newAssessment = clone(assessment)
+    newAssessment.conducted_by_last_name = event.target.value
+    const wrapper = shallow(<ConductedBy {...props} />)
+    wrapper.instance().handleConductedByLastNameChange(event)
```
wmitchell-cambria: Instead of calling the internal method using `instance()`, it's more robust to find the prop on the child and call that:

```
wrapper.find('#conducted-by-first-name').props().onChange(event)
```

This style continues working even with internal refactorings, such as renaming the method or wrapping it in a helper function. It also tests that the prop is wired up correctly, so we don't have to add the expectation separately on line 31.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByNameField.js
```diff
@@ -0,0 +1,53 @@
...
+  onChange: () => {},
+  disabled: false,
+  errorMessage: '',
+}
+
+ConductedByNameField.propTypes = {
+  disabled: PropTypes.bool,
+  errorMessage: PropTypes.string,
+  id: PropTypes.string.isRequired,
+  onChange: PropTypes.func,
```
wmitchell-cambria: Is there a situation outside of tests where we would want this to be undefined? Tests can easily provide a default callback, if that is the only scenario.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Select } from '@cwds/components'
+import { ConductedByRoleOptions } from '../AssessmentHelper'
+import { findSelectOptionByValue } from '../../../util/common'
+
+const ConductedByRole = ({ value, onChange, disabled }) => {
+  const selectedOption = findSelectOptionByValue(value, ConductedByRoleOptions)
+  return (
+    <div className={'domain-comment-block'}>
```
wmitchell-cambria: Is the extra `div` necessary? I'm guessing it is not needed.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js
```diff
@@ -0,0 +1,39 @@
...
+ConductedByRole.defaultProps = {
+  value: '',
+  placeholder: undefined,
+  onChange: () => {},
+  disabled: false,
+}
+
+ConductedByRole.propTypes = {
+  disabled: PropTypes.bool,
+  onChange: PropTypes.func,
```
wmitchell-cambria: Same question here. Do we ever want this to be unspecified outside of tests?

app/javascript/Application/components/Assessment/AssessmentHelper.js
```diff
@@ -39,7 +55,11 @@ function validateAssessmentDtoFields(assessmentDto) {
     validateAssessmentEventDate(assessmentDto.person.dob, assessmentDto.event_date) &&
     isDefined(assessmentDto.assessment_type) &&
     isDefined(assessmentDto.completed_as) &&
-    isDefined(assessmentDto.has_caregiver)
+    isDefined(assessmentDto.has_caregiver) &&
+    isDefined(assessmentDto.conducted_by_first_name) &&
```
wmitchell-cambria: Tests for the changes here?

app/javascript/Application/util/common.js
```diff
@@ -46,3 +46,6 @@ export const basePath = process.env.CANS_BASE_PATH || '/cans'
 export const getPageRoute = () => {
   return basePath === '/' ? document.location.pathname : document.location.pathname.substring(basePath.length)
 }
+export const findSelectOptionByValue = (value, selectOptions) => {
+  return value ? selectOptions.find(option => option.value === value) : undefined
```
wmitchell-cambria: It doesn't look like we test this function with `value` as undefined/falsy. Is that a situation that can come up? We should either add a test for it or remove the guard.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
DmitriyRudenko: clone is used everywhere where in cans app where "update assessment" event happens including "handleConductedBy...Change" methods which were there before as well. So if we consider to do corresponding refactoring then we have to remove/replace clone everywhere. This kind of refactoring is a good candidate for tech debt task.

wmitchell-cambria: Respectfully disagree. Clone was the direct cause of some of the performance issues that we addressed yesterday, and it can be replaced piece by piece to see improvements. We certainly should avoid adding _more_ instances of it.

When new information arises, we cannot keep doing it the same way just because that is how it used be done.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
DmitriyRudenko: Same thing here. onAssessmentUpdate() callback requires assessment object as parameter. And this approach used everywhere as well, so refactoring should affect all other similar places in entire app. It good candidate for tech debt task.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
DmitriyRudenko: Implementation is based on AC, link: https://app.goabstract.com/projects/6a9199d0-4cc7-11e8-8f6e-a5a02644863c/branches/f3529d40-c757-11e8-a5c2-196c475d3a45/collections/504e1316-de72-411a-aec1-86482067c151/presentation

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js
DmitriyRudenko: those functions are not props

wmitchell-cambria: The instance function is passed to a child as a prop. It is asserted to be so on line 31:
```expect(firstNameInput.props().onChange).toBe(wrapper.instance().handleConductedByFirstNameChange)```
This is a pretty common pattern that I comment about a lot. Hit me up on Slack if you want to discuss the details.

File: app/javascript/Application/components/Assessment/AssessmentHelper.js (hunk 10 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentHelper.js]: there are a lot of tests related to assessment completion process in AssessmentContainer.test.js like describe('Complete button'... etc, so I have modified assessment mock objects respectively

File: app/javascript/Application/util/common.js (hunk 11 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/util/common.js]: we have two tests: when option is present and when it isn't present. the condition: "value ?" is just added for performance reasons(to not iterate over the list if it isn't necessary.

Comment (wmitchell-cambria) [reply app/javascript/Application/util/common.js]: Since the list is controlled on our end, how often will it be able to skip iterating the list?

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js (hunk 9 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js]: done

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByNameField.js (hunk 7 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByNameField.js]: done

**Comment by DmitriyRudenko** [comment]
@wmitchell-cambria  backend PRs: https://github.com/ca-cwds/cans-api/pull/259, https://github.com/ca-cwds/cwds-data-model/pull/314

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js (hunk 8 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByRole.js]: done

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js (hunk 1 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js]: done

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js (hunk 2 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js]: done

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js (hunk 3 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js]: done

File: app/javascript/Application/util/common.js (hunk 11 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/util/common.js]: removed

**Comment by wmitchell-cambria** [comment]
Changes so far look good, thanks. There's one outstanding question on the function testing, and then the field redesign as well.

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js (hunk 6 above)
Comment (DmitriyRudenko) [reply app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js]: done

File: app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.test.js [hunk 12]
```diff
@@ -67,7 +67,10 @@ describe('Assessment Conducted by', () => {
     const newAssessment = clone(assessment)
     newAssessment.conducted_by_first_name = event.target.value
     const wrapper = shallow(<ConductedBy {...props} />)
-    wrapper.instance().handleConductedByFirstNameChange(event)
+    wrapper
+      .find('#conducted-by-first-name')
+      .props()
+      .onChange(event)
```
wmitchell-cambria: 👍

wmitchell-cambria:
Offline, we chatted with Chilton to get a new design that satisfies accessibility for this fieldset. I would have liked that to be in the same story (since all stories should meet our quality bar, including accessibility). It's not, though, and it was split out to here: https://osi-cwds.atlassian.net/browse/CANS-1183

I'll do one last pass over the PR to make sure it is good to go in the meantime. CANS-1183 should be following this one up in the near future.

app/javascript/Application/components/Assessment/style.sass
```diff
@@ -399,3 +399,33 @@ $item-confidential-label-height: 6.05rem
 .-redacted-level-radio-group
   padding-left: 2rem
   padding-top: 1rem
+
+//*********************************
+//  Form fields
+//*********************************
+
+.conducted-by-name-field-err
+  border-color: $error-red !important
```
wmitchell-cambria: Do we have a story in the backlog yet to update these styles once the Component Library releases form controls? Validation and error styles should definitely be shared across services, and it will help us avoid `!important` rules like this. If we don't have one, can we please create one and mark it as blocked by a relevant CL story? (It doesn't necessarily have to be marked for the current fix version; I just don't want us to lose track of it.)

DmitriyRudenko: Form fields still not available in component lib. Once we will switch to them those styles will be removed

wmitchell-cambria: Is there a story for that yet, though?

DmitriyRudenko: PR: https://github.com/ca-cwds/design-system/pull/477

DmitriyRudenko:
jenkins retest this please

PR #681, opened 2019-04-26
Cans 1178 fix production build issues

TylerOSI:
## JIRA
[CANS-1178](https://osi-cwds.atlassian.net/browse/CANS-1178)

## Description
This fixes a problem where some assets that are fetched by the browser are not properly prefixed with the `/cans/` prefix.

I also removed a hack for fetching an icon that is not needed anymore and can be a pain point as the SHAs may change during compilation.

## Tests

## Types of changes

## Checklist:


wmitchell-cambria:
Retest this please.

TylerOSI:
promoting this to a PR because it looks like this is the only way to currently fix the problem. There is an environment variable to set this prefix, but `process.env` does not have this variable so there is no way to access it in `config/webpack/production.js`

There may be [another way](https://github.com/rails/webpacker/issues/22) to fix this, but it involves setting the base url (which we do not want to do).

The caveat is that if we decide to update `CANS_BASE_PATH` we have to change this hard coded value.

PR #684, opened 2019-04-26
CANS-1182 Use ToolTips for Domains on Assessment Page

nasar-cambria:
https://osi-cwds.atlassian.net/browse/CANS-1182

## Description
ToolTips are used for domains and pop over size is updated so that CANS application is consistent in way it displays information about the application without flickering popover

**NOTE:** As confirmed by design team there would not be additional sizes for infoTip popover. We are going to use toolTip and customize the popover to be little wider so that it doesn't flicker and show the complete information to the user. The icons for tooltips will also be customized to match infoTip icons as shown in CANS-Summary. Also, ToolTips doesn't accept className props so I have used inline styles for this component.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Assessment/Domain.test.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Button, Icon, UncontrolledInfotip } from '@cwds/components'
```
LeonidMarushevskyi: There is no way to customize UncontrolledInfoTip?

app/javascript/Application/components/Assessment/DomainPanelSummary.js
```diff
@@ -20,9 +20,22 @@ class DomainPanelSummary extends React.PureComponent {
...
-      <UncontrolledInfotip id={`domain-${index}`} placement="top">
-        <PopoverBody>{description}</PopoverBody>
-      </UncontrolledInfotip>
+      <UncontrolledTooltip
+        style={{
+          minWidth: '32rem',
+          textAlign: 'left',
+          padding: '1.2rem',
+          fontSize: '1rem',
+          opacity: '1.5!important',
```
zliuwd: I don't really get what "opacity:1.5" here stand for, but the PR looks safe enough to merge.

wmitchell-cambria: Doesn't opacity range between 0 (completely transparent) and 1 (completely opaque)? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity), any value outside of the range is clamped to the nearest limit, so 1.5 should behave the same as 1. If it does not, then it is likely a browser bug.

app/javascript/Application/components/Assessment/Domain.test.js
nasar-cambria: Unfortunately no. That’s why it was decided to bring back toolTip after discussing with Robert Chilton.

app/javascript/Application/components/Assessment/DomainPanelSummary.js
nasar-cambria: You are right opacity is between 0 and 1 (1 being 100%) but for some reason it was still looking a bit transparent after making it to 1. So, I spoke to Robert from the design team about this and he confirmed that this will be corrected in the component itself but there is no time line provided for that update. I intended to leave it as is until that update is shipped by team CL.

nasar-cambria:
Re-test this please

PR #685, opened 2019-04-27
Small unit test refactors

wmitchell-cambria:
## JIRA

No story, just refactoring.

## Description

This started out as an effort to speed up some of our unit tests, primarily by replacing `mount` with `shallow` where possible. It turns out that this was helpful, but that Jest was spending a lot of time in startup. (If you want to fix that, using `yarn test -i` seems to help.)

In any case, I managed to replace some of the mounts with shallow and refactor the test suite to be a little less verbose (and at the same time more clear about what was under test).

When reviewing, please be sure to expand the `Domain.test.js` file. GitHub hides large diffs by default, but it is important.

## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/common/DateField.test.js
```diff
@@ -107,21 +107,23 @@ describe('DateField', () => {
   })
 
   describe('when value is null, emptystring, or undefined', () => {
-    describe('with time', () =>
-      [null, '', undefined].map(value => {
+    describe('with time', () => {
+      new Array(...[null, '', undefined]).map(value => {
```
TylerOSI: why? just curious.

wmitchell-cambria: Oh, right...um... so the story here...

There was a warning printed somewhere along the way that `describe` blocks that return a value are deprecated and will start throwing an error in the next version. So, I added the `{}`.

Now that the arrow function was a code block, Prettier wanted the line to start with a semicolon. Statements starting with an array literal can often be ambiguous, so Prettier wanted it to be:

```javascript
describe('with time', () => {
  ;[null, '', undefined].map(...)
})
```

Once I did that, the Eslint rule for `no-extra-semi` was triggered, asking the semicolon to be removed. (Note that in this case, Eslint is 'right'. There is no actual ambiguity in how the statement should be parsed.)

So, short story:  Avoiding a Jest deprecation while satisfying linting.

(For fun, adding an extra line does show the ambiguity)
```javascript
describe('with time', () => {
  const foo = bar
  [null, '', undefined].map(...)
})
// ...is parsed as...
describe('with time', () => {
  const str = bar[undefined].map(...)
})
```

wmitchell-cambria: I'll push a small commit to clean this up. I don't like the spread operator.

PR #689, opened 2019-04-29
Cans 1030 test fix

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


DmitriyRudenko:
retest this please

spec/page_objects/assessment_form.rb
```diff
@@ -230,6 +228,15 @@ def fill_conducted_by_last_name_field(text)
   )
 end
 
+def set_value_for_text_field(field, text)
+  field.click
+  sleep 2
```
wmitchell-cambria: Is this just a temporary fix, or a proposed final solution?

wmitchell-cambria:
It looks like the fields are causing a large amount of re-rendering on every keystroke. This would definitely explain why Capybara has issues entering in values quickly. We should fix the rendering issues themselves instead of adding sleeps to the test files.

Render Highlighting

spec/page_objects/assessment_form.rb
DmitriyRudenko: temp fix

DmitriyRudenko: definitely something wrong with capybara...

DmitriyRudenko:
@wmitchell-cambria thanks for addressing this, I'll take a look, I think main reason is onAssessmentUpdate callback

PR #692, opened 2019-04-30
Assessment prop is removed from conducted by component

DmitriyRudenko:
## Tests

## Types of changes

## Checklist:


LeonidMarushevskyi:
Jenkins retest this please

wmitchell-cambria:
I'll take a look this morning.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -1,73 +1,58 @@
...
-    assessment.conducted_by_last_name = event.target.value
-    this.props.onAssessmentUpdate(assessment)
-  }
-
-  handleConductedByRoleChange = event => {
-    const assessment = { ...this.props.assessment }
-    assessment.conducted_by_role = event.value
-    this.props.onAssessmentUpdate(assessment)
-  }
+const ConductedBy = ({ firstName, lastName, role, onChange, disabled }) => {
```
wmitchell-cambria: Stateless functional components will always re-render, even if their props don't change. We can make this a PureComponent. (This will probably trigger a Lint rule...see my next comment.)

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -1,73 +1,58 @@
...
+    <form>
+      <fieldset>
+        <legend className={'assessment-form-header-legend'}>Assessment Conducted By</legend>
+        <Row className={'assessment-form-header-inputs'}>
+          <Col sm={4}>
+            <ConductedByNameField
+              id={'conducted-by-first-name'}
+              label={'First Name *'}
+              value={firstName}
+              onChange={onChange}
```
wmitchell-cambria: Instead of passing the `onChange` in directly, let's keep the field-based helper methods here (same names as before). Each one can take in the value and add the correct field. Like:

```javascript
handleConductedByFirstNameChange = name => {
  this.props.onChange('conducted_by_first_name', name)
}
```

By creating these methods here, we can avoid the need for `fieldName` as a prop. Plus, the helper methods will avoid the Lint rule for prefer-functional-component.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
```diff
@@ -1,73 +1,58 @@
...
+              id={'conducted-by-last-name'}
+              label={'Last Name *'}
+              value={lastName}
+              onChange={onChange}
+              disabled={disabled}
+              fieldName={'conducted_by_last_name'}
+            />
+          </Col>
+          <Col sm={4}>
+            <ConductedByRole value={role} onChange={onChange} disabled={disabled} />
```
wmitchell-cambria: Let's do the same thing with this onChange, for consistency.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByNameField.js
```diff
@@ -1,39 +1,46 @@
...
-        </div>
-      )}
-      {!disabled && (
-        <span className={`form-field-length${showError ? '-err' : ''}`}>{`${
-          value.length
-        }/${ConductedByNameMaxLength}`}</span>
-      )}
-    </div>
-  )
+class ConductedByNameField extends Component {
```
wmitchell-cambria: This component (and `ConductedByRole`) can extend from `PureComponent` instead of `Component`. `PureComponent` should generally be preferred.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedByNameField.js
```diff
@@ -1,39 +1,46 @@
...
+  handleConductedByNameChange = event => {
+    this.props.onChange(this.props.fieldName, event.target.value)
```
wmitchell-cambria: Based on the previous changes, this will become `this.props.onChange(event.target.value)`. `fieldName` prop will go away.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ConductedBy.js
DmitriyRudenko: we have same component both for first and last names, so this is not an option

wmitchell-cambria: Wasn't that how it was set up before, though?
```
onChange={this.handleConductedByFirstNameChange}
onChange={this.handleConductedByLastNameChange}
```

DmitriyRudenko: I see, ConductedBy  should be PureComponent, not ConductedByNameField

wmitchell-cambria: Both should be.

wmitchell-cambria:
Oh, good. I still think the remaining changes with the fieldName and callbacks would be good, but not critical. My reasoning is that by passing fewer props to the field components, they stay simpler and easier to test fully. (This is a variant of 'Composition over Customization'.) If you have the appetite for making that change, go for it. Otherwise, I'll approve either way.

DmitriyRudenko:
retest this please

PR #696, opened 2019-05-01
AssessmentFormHeader refactor

wmitchell-cambria:
## Description

This one came out a little bit bigger than I anticipated. This tackles the AssessmentFormHeader, refactoring with one goal in mind:  Making as few things as possible depend on the full `assessment` object. This was not actually too difficult...just tedious. Helper methods were split into their own components, and PureComponents were used where possible to avoid rendering excessively. The most important affect here is that as you type into the ConductedBy fields, as little as possible is re-rendered, so those fields behave with much less lag.

There was an ugly merge on this one as well. I think I got all of @abelayt's recent a11y changes, but extra eyes would be appreciated.

## Tests

## Types of changes

## Checklist:


DmitriyRudenko:
How about to remove "retry" workaround from regression tests for "conducted by first/last name" fields? file: spec/page_objects/accessment_form.rb, lines: 207-238

wmitchell-cambria:
We should be able to, but I wanted to separate it into its own PR. I think these refactors are generically good to have. I _think_ the retry workaround will not be necessary after this, but I am not 100% sure. I would rather test them separately.

Also, Jenkins: retest this please.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientDateOfBirth.js
```diff
@@ -0,0 +1,17 @@
+import PropTypes from 'prop-types'
+import { isoToLocalDate, isValidDate } from '../../../util/dateHelper'
+
+const ClientDateOfBirth = ({ dob, isEstimated }) =>
+  isValidDate(dob) ? `DOB: ${isoToLocalDate(dob)}${isEstimated ? ' (approx.)' : ''}` : ''
```
TylerOSI: nested ternary is not a lint thing? It probably would be easier to read if the approx postfix was separated out into its own function or varable:

```
  isValidDate(dob) ? `DOB: ${isoToLocalDate(dob)}${approxPostfix}` : ''"
```

TylerOSI: also render `null` instead of empty string, since rendering null produces nothing, but rendering empty string produces an empty span

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
```diff
@@ -0,0 +1,65 @@
...
+
+  render() {
+    const { dob, isEstimatedDob, firstName, middleName, lastName, suffix } = this.props
+    return (
+      <div>
+        {firstName && lastName ? (
+          <div>
+            <div className={'card-title-block'}>
+              <span id={'child-name'}>
+                {formatClientName({ first_name: firstName, middle_name: middleName, last_name: lastName, suffix })}
```
TylerOSI: if the default name props are used this will render an empty span, or perhaps an empty span inside an empty span?

app/javascript/Application/components/Assessment/AssessmentFormHeader/DateAndTemplate.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Row, Col, Label } from 'reactstrap'
```
denys-davydov: The PR is a good place to get rid of the reactstrap components when possible

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
```diff
@@ -0,0 +1,65 @@
...
+import { formatClientName } from '../../Client/Client.helper'
+import { calculateDateDifferenceInYears, isValidDate } from '../../../util/dateHelper'
+import ClientDateOfBirth from './ClientDateOfBirth'
+
+class ClientNameTitle extends React.PureComponent {
+  formatClientAge(dob) {
+    return isValidDate(dob) ? `${calculateDateDifferenceInYears(dob, this.getCurrentDate())} years old` : ''
+  }
+
+  getCurrentDate() {
```
denys-davydov: It could be good to use `dateHelper#today()` function instead of having a new function and introducing a `moment` library dependency for the component.

wmitchell-cambria:
retest this please

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
denys-davydov: First we are switching from underscore_lowercase to camelCase and then we have to switch back :(

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import moment from 'moment/moment'
+import { formatClientName } from '../../Client/Client.helper'
+import { calculateDateDifferenceInYears, isValidDate } from '../../../util/dateHelper'
+import ClientDateOfBirth from './ClientDateOfBirth'
```
denys-davydov: Would it be reasonable to import `styles.sass` here as well as in other new components that are using styles?

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientDateOfBirth.js
wmitchell-cambria: I agree. The linting is probably not picking up on it because it is within a template string.

However... this code is not new, just moved: https://github.com/ca-cwds/cans/pull/696/files#diff-aebe384a886f79699b1d6e6756b46411L93 For exactly the reason you mentioned (to avoid potential spacing issues across browsers), I'd like to keep it as is for now and fix it in a new PR if necessary.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
wmitchell-cambria: If the default name props are used, this entire block will not render. It is guarded by `firstName` and `lastName` being truthy.

app/javascript/Application/components/Assessment/AssessmentFormHeader/ClientNameTitle.js
wmitchell-cambria: Possibly... I'm not sure. I didn't really think about styles at all. I think it could make sense to split them into smaller stylesheets, perhaps one for each component, and _then_ include the styles.

wmitchell-cambria:
Assuming this passes test, my plan is to merge it and then follow up with a new PR for the remaining code review comments. (That second PR will be much smaller and less prone to merge conflicts.)

PR #668, opened 2019-04-22
Upgrade webpack and dependencies

TylerOSI:
## JIRA
No direct story

## Description
building the docker image was taking 10 minutes on every PR build. This should speed it up. old versions of webpacker use [extract-text-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) which was slowing down compile time. Newer versions of webpacker use [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin) instead.

... so i upgraded webpacker and a whole bunch of related dependencies. The application is running with styles in place, and tests seem to be passing so far.

## Tests

- [ ] I used TDD to develop the feature
- [ ] I have included unit tests
- [ ] I have included new acceptance tests or modified existing ones
- [ ] I have included other tests

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (No behavioral changes)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I ran all my tests locally which were green.
- [ ] My code adds no new issues to Code Climate
- [ ] I ran all the linters for the project.
- [ ] My code is in a stable state ready to be deployable, but not necessarily complete.
- [ ] I promise on my honor as a CWDS developer to ensure I don't break the build (or bring donuts if I leave things broken), to check linters, use best practices, to leave the code in better shape than I found it, and to have a good day.

PR #700, opened 2019-05-01
CANS 1179 rebuild base test image

TylerOSI:
## Description
This fixes some base image dependencies and upgrades chrome to the version 74.

upgraded:
* chrome
* chrome-driver
* selenium-webdriver
* capybara

notes:
* tests start a new browser session for each test case
* caregiver name could not be set to empty string for some reason

## Tests

## Types of changes

## Checklist:


TylerOSI:
Jenkins test this please

TylerOSI:
related to #668

TylerOSI:
related to #674

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -12,7 +12,7 @@
   current_date = Time.now.getlocal.strftime('%m/%d/%Y')
   assessment_link = 'CANS Assessment Form'
 
-  before(:all) do
+  before(:each) do
```
denys-davydov: Can't find what has changed that we need to login before each scenario.

TylerOSI:
Jenkins test this please

spec/regression/case_worker_scenarios_spec.rb
TylerOSI: I can do some research to try to find out. It started after i updated chrome, chrome driver, and related dependencies.

TylerOSI: seems like this is due to how [Capybara is hooked into Rspec](https://github.com/teamcapybara/capybara/blob/4da9195870d7782bce3d3b0db27e0f793ea24343/lib/capybara/rspec.rb#L18). it clears the session after each test run, which brings the user back to login. We can monkey patch this code by removing `require capybara/rspec` and configuring Rspec to connect to Capybara ourselves. Though there are performance benefits for doing that, I don't think its worth it.

denys-davydov:
Retest this please

denys-davydov:
The user story is postponed so closing the PR

PR #703, opened 2019-05-03
CANS-1193 Fixing Print iframe was rendered as a "visible" element

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1193
## Description

- #### Making Print iframe container be rendered as an "invisible" element
The main reason which caused this bug is iframe was rendered with height and width, reverse them back to "zero" as the old Print did. Technically the size of the iframe should not impact the actual print output, because our print is focused on the content inside the iframe and we already pre-insert some size setting into the iframe before the print target component loading.
- #### Putting back Client page print style to its own folder
### Why?
1. For now, Print should be a pure common tool, the style related setting for it should be only the basic border thickness and the style for the reusable components inside PrintUtil

2. We should put all style setting which related to "target print component" (here is PrintClient.js) in their own folder because they are "inline style objects". If these objects were spread to everywhere it would be very difficult to manage and maintain by the devs who will work on them later. 

3. I believe the reason we created PrintStyle.js file in the root Print folder is we want to reuse it. But based on the points upwards and the content of PrintStyle File, we will find all the setting in PrintStyle.js are related to one or more specific detail of "Print target component" (PrintClient.js), so they should be put back to their own folder.
## Tests

## Types of changes

## Checklist:


app/javascript/Application/components/Print/Print.js
```diff
@@ -15,9 +15,11 @@ class Print extends React.Component {
...
         id={printFrameId}
         title={printFrameId}
-        style={{ height: '100%', width: '100%' }}
-        height="100%"
-        width="100%"
+        style={{
+          height: 0,
+          width: 0,
+          position: 'absolute',
+        }}
```
zliuwd: The reason using inline style here is: want to make sure the unit test could cover every single value of style setting.

PR #708, opened 2019-05-10
CANS-1186 Comparison graph regression test

zliuwd:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1186

## Tests

## Types of changes

## Checklist:


zliuwd:
retest this please

spec/regression/case_worker_scenarios_spec.rb
```diff
@@ -643,4 +662,56 @@ def go_forward
   def reload_page
     page.evaluate_script('window.location.reload()')
   end
+
+  def legend_text_to_date_info(text)
+    date = Time.strptime(text.slice(0..9), '%m/%d/%Y')
+    if text.length == 10
+      [date, 0]
+    else
+      [date, text.slice(10..-1).delete('()').to_i]
```
DmitriyRudenko: how about to `text.slice(11..-2).to_i` ?

spec/page_objects/assessment_comparison.rb
```diff
@@ -0,0 +1,24 @@
...
+  elements :comparison_outter_rating_box, 'div.item-rating'
+  elements :graph_legend, 'span.recharts-legend-item-text'
+  element :age_switch_0to5, 'button#age-0-5-button'
+  element :age_switch_6to21, 'button#age-6-21-button'
+  element :graph_horizontal_axis, 'g.recharts-xAxis'
+  elements :graph_horizontal_axis_ticks, 'text.recharts-cartesian-axis-tick-value'
+  elements :graph_bars, 'g.recharts-bar-rectangle'
+  element :graph_bar_tool_tip, 'ul.recharts-tooltip-item-list'
+  elements :graph_bar_tool_tip_values, 'span.recharts-tooltip-item-value'
+
```
DmitriyRudenko: I think below methods are redundant

spec/regression/case_worker_scenarios_spec.rb
zliuwd: Sorry for late reply Dmitriy, this is the way to catch the suffix of the dates, e.g. it will catch "(2)" from  01/01/2019(2)

spec/page_objects/assessment_comparison.rb
zliuwd: If we check the rechart github we will find this:
```return (
            <li className="recharts-tooltip-item" key={`tooltip-item-${i}`} style={finalItemStyle}>
              {isNumOrStr(name) ? <span className="recharts-tooltip-item-name">{name}</span> : null}
              {isNumOrStr(name) ? <span className="recharts-tooltip-item-separator">{separator}</span> : null}
              <span className="recharts-tooltip-item-value">{value}</span>
              <span className="recharts-tooltip-item-unit">{entry.unit || ''}</span>
            </li>
          );```
so, 1. `elements :graph_bars, 'g.recharts-bar-rectangle'` will provide a position to "hover"
then, 2. the tooltip col will show `element :graph_bar_tool_tip, 'ul.recharts-tooltip-item-list'`
finally, 3. we will only catch the value per assessment date in the tool tip ` elements :graph_bar_tool_tip_values, 'span.recharts-tooltip-item-value'`

Therefore, these three lines are essential

spec/regression/case_worker_scenarios_spec.rb
zliuwd: For the case that we will only have 5 assessments tops, yes! `text.slice(11..-2).to_i` will work, but `text.slice(10..-1).delete('()').to_i` will work for the case there is a two digit number inside the () also. 
You got neat logic man!

PR #715, opened 2019-06-26
CANS-1266. Feature flag for the Domain Total functionality

denys-davydov:
## JIRA
https://osi-cwds.atlassian.net/browse/CANS-1266

## Description
Since there is no final decision made if the Domain Total feature will be fully removed or reworked in future we suggest introducing a temporary feature flag that will hide the functionality from the user but the source code will still be available for possible future changes. This approach will place us in a very flexible position even if eventually no changes will be requested.
The Domain Total is not stored in the database and calculated on the fly on the front-end, so the feature flag should be applied to the front-end application only.

This PR introduces **DOMAIN_TOTAL_FEATURE_ENABLED** feature flag, that can turn the Domain Total functionality on/off.

## Tests

## Types of changes

## Checklist:


denys-davydov:
> @denys-davydov do we intend to add this information to README file as well ?

Sure @nasar-cambria it is added, you may see by this link

package.json
```diff
@@ -28,6 +28,7 @@
     "axios": "^0.18.0",
     "bootstrap": "^4.3.1",
     "classnames": "^2.2.5",
+    "dotenv": "^8.0.0",
```
LeonidMarushevskyi: Should we run license update?

app/javascript/Application/util/env.js
```diff
@@ -0,0 +1 @@
+export const isDomainTotalFeatureEnabled = () => process.env.DOMAIN_TOTAL_FEATURE_ENABLED !== 'false'
```
vladimirPe: Maybe in our case better to have DOMAIN_TOTAL_FEATURE_DISABLED, because we already have that functionality and the default behavior is to show domains totals.
And I would use `new Boolean(process.env.DOMAIN_TOTAL_FEATURE_DISABLED)  !== true`

denys-davydov: I believe there is a best practice to name variables in a positive manner rather than a negative to avoid double negations which complicate an understanding of the code.

package.json
denys-davydov: Updated licenses

